PR #39, opened 2018-06-08
Add profile example

justinaray:
Demos fetching API Data
Demos showProfile sdk functionality
Add an icon for the new profile app


examples/profile.html
```diff
@@ -0,0 +1,307 @@
...
+             */
+            function renderUserProfile(profileRootEl, profileData) {
+                let linkEl = profileRootEl.querySelector('.profile-link');
+
+                if (profileLinkListener) {
+                    linkEl.removeEventListener('click', profileLinkListener);
+                    profileLinkListener = null;
+                }
+
+                if (profileData.id) {
```
justinaray: Should put this check inside the click listener so we get some logging on click and can intercept the navigation.

examples/profile.html
```diff
@@ -0,0 +1,307 @@
...
+                    });
+                }
+
+                let placeholderPicEl = profileRootEl.querySelector('.placeholder-pic');
+                let picEl = profileRootEl.querySelector('.profile-pic');
+                if (bestImage.size) {
+                    setHidden(placeholderPicEl, true);
+
+                    picEl.setAttribute('src', bestImage.uri);
+                    if (profileData.name) {
```
justinaray: Need an else here since the idea is to have this functionally set the provided element. ... even though it's not a component and wont be reused.

examples/profile.html
```diff
@@ -0,0 +1,307 @@
...
+                if (oauthParams && oauthParams.access_token && oauthParams.state) {
+                  // OAuth2 spec dictates this encoding
+                  // See: https://tools.ietf.org/html/rfc6749#appendix-B
+                  let stateSearch = unescape(oauthParams.state);
+                  result = extractParams(stateSearch).pcEnvironment;
+                }
+              }
+
+              if (!result && window.location.search) {
+                result = extractParams(window.location.search.substring(1)).pcEnvironment;
```
justinaray: This will technically result in undefined.  Need to modify as the doc states it will return as null.

PR #42, opened 2018-06-13
Add Frankfurt (.de tld) to list of regions

justinaray:
Add frankurt region
Update valid region test to fail assertion rather than throw js error


justinaray:
@KevinGlinski ... still need to run one more test in the morning before merging.

justinaray:
Tested and it works.  Ready to go from my standpoint.

PR #46, opened 2018-09-11
Feature/show interaction details

griffind:


src/modules/ui.js
```diff
@@ -60,6 +60,18 @@ class CoreUiApi extends BaseApi {
     hideHelp() {
         super.sendMsgToPc('hideHelp');
     }
+
+    /**
+     * Show an interaction by ID.
```
justinaray: Thinking of moving this out to a conversations.js module?  This will mirror the platform api which will facilitate users already familiar with it.  Also, the showUserDetails is already in a separate users.js module so it would be consistent there.

Thoughts?

src/modules/ui.js
```diff
@@ -60,6 +60,18 @@ class CoreUiApi extends BaseApi {
...
+     *
+     * @example
+     * myClientApp.coreUi.showInteractionDetails('af2ef59d-9bc5-4436-8738-97c04869c81c');
+     *
+     * @since 1.1.0
+     */
+    showInteractionDetails(conversationId) {
```
justinaray: After moving this out to a separate module, would you mind adding a test for it?  You can follow lifecycleSpec.js to mock up the `_myParent` prop and assert the message is sent.

examples/interactionsDemo.html
```diff
@@ -0,0 +1,526 @@
...
+                    <th scope="col">Wrap Up</th>
+                </tr>
+            </thead>
+            <tbody id="tableBody">
+            </tbody>
+        </table>
+        <section class="reloading fas fa-spinner fa-spin fa-3x hidden">
+        </section>
+
+        <div class="form-group conversationID">
```
justinaray: Can we add a header somewhere to indicate that this bit is a dev/testing tool?

examples/interactionsDemo.html
```diff
@@ -0,0 +1,526 @@
...
+        <h1>Show Interaction Details (Past 30 days)</h1>
+
+        <section class="authenticating hidden">
+            Authenticating to PureCloud <span class="fas fa-spinner fa-spin"></span>
+        </section>
+        <section class="failure hidden">
+            Sorry, an error has occurred.
+        </section>
+
+        <div class="topGroup">
```
justinaray: Ideally, this section would only show once the initial setup and auth is done.  It can/should be shown in the fetch failure case though so that the user can retry.

examples/interactionsDemo.html
```diff
@@ -0,0 +1,526 @@
...
+                        if (!authenticated) {
+                            setErrorState('Failed to Authenticate with PureCloud');
+                        }
+                    });
+
+                /**
+                 * Makes the request to the API to get the current users interactions for
+                 * the past 30 days.
+                 */
+                function makeSDKRequest() {
```
justinaray: maybe pass the userId in here in this function so you don't have to rely on this?  You can close over the userId when building the reload click handler too so you won't have to deal with this at all.

examples/interactionsDemo.html
```diff
@@ -0,0 +1,526 @@
...
+                                        "type": "dimension",
+                                        "dimension": "userId",
+                                        "operator": "matches",
+                                        "value": this.profileData.id
+                                    }
+                                ]
+                            }
+                        ]
+                    };
+
```
justinaray: Do you need to disable and/or hide items here so it works in both the initial load and reload case?

src/modules/ui.js
griffind: Yea we should keep it consistent with the api.

examples/interactionsDemo.html
griffind: Can you clarify what you mean here? What items and what are you concerned might break because of them?

justinaray: iirc, I was thinking that you should "reset" the page state to the initial state here so you make sure it's in a correct state by the end of the function.  for example, maybe clear and hide the table and error message.  Not sure off the top of my head if there are more.

Now, it's also possible that you could handle these cases below.  For instance, only clear the error state in the success handler and only clear and hide the table in the error case.  Although this opens you up to get in a bad state if the function fails fast or errors out for some reason now or in the future.

Basically advocating for a "functional" style ui update here.  With that caveat that it's not functional at all since it has side effects ;)

Of course, if all of the cases are handled, success -> failure, failure -> success, failure -> failure, success -> success, I don't really mind how you get there.

examples/interactionsDemo.html
```diff
@@ -0,0 +1,517 @@
...
+                        }
+
+                        if (checkPermission(permissions, 'routing:wrapupCode:view')) {
+                            canViewWrapUpCodes = true;
+                        }
+
+                        setHidden(authenticatingEl, true);
+
+                        $('#reloadButton').on('click', function () {
+                            $('.failure').text('');
```
griffind: If I understand what you are saying about the error state correctly, does this line and line 228 handle all cases correctly by hiding the failure message and removing the table body? This would be triggered on every page load and button reload.

examples/interactionsDemo.html
```diff
@@ -0,0 +1,532 @@
...
+                            $(this).attr('disabled', '');
+                            setHidden(reloadingEl, false);
+                            $('#tableBody tr').remove();
+
+                            // 30 days in the past
+                            let today = new Date();
+                            let today30 = (new Date(new Date().setDate(today.getDate() - 30))).toISOString();
+
+                            return (function() {
```
justinaray: Any reason to return here?  I don't think jquery will care about anything except truthy/falsy here and would only prevent the default.  Not sure if this was an attempt to block clicks based on the resolution of the promise for instance.

Since the refresh button is only re-enabled on the promise resolution/rejection, we shouldn't need to block otherwise.

Could debounce the click handler, but probably overkill for a demo app.

examples/interactionsDemo.html
```diff
@@ -0,0 +1,532 @@
...
+                                                    return true;
+                                                }
+                                            });
+
+                                            if (canViewWrapUpCodes) {
+                                                conversation.participants.forEach(function (participant) {
+                                                    participant.sessions.forEach(function (session) {
+                                                        session.segments.some(function (segment) {
+                                                            if (segment.hasOwnProperty('wrapUpCode') && typeof segment.wrapUpCode != 'undefined' && segment.wrapUpCode != '') {
+                                                                new platformClient.RoutingApi().getRoutingWrapupcode(segment.wrapUpCode)
```
justinaray: This nested promise worries me a bit from a memory leak perspective.   If there's a bunch of reloads with wrap-up codes, will those references start to stack up and fail to remove the row listeners; leading to a leak?

Now, this is a demo app so it would take a lot of atypical usage to start to accumulate memory and won't be used in production at that.  Also, the fix would probably be to decouple data acquisition from markup building (and cache routing code results) so that would probably balloon this effort out more than we need it to.

examples/interactionsDemo.html
justinaray: Yeah, I think only adding the click listener after successful auth and user retrieval limits some of the edge cases I was seeing before.  Now, the error text will only be used for reloads after initial success, so there shouldn't be any polluting state.

We're definitely pushing the limits of what we can do in these demo apps by keeping it "simple" with no build, dependencies, etc.  Reminds me of the good 'ol days of jquery spaghetti!

examples/interactionsDemo.html
griffind: TBH, I do not recall why I returned here but as you said it is not needed. I can pull it out.

examples/interactionsDemo.html
griffind: Perhaps? When I tear down the table body on line 241 does that not free up that memory and remove the row listeners? I took some memory snapshots in chrome and didn't see any nodes without direct references to JS and detached from the DOM tree. It seemed to me that the table was getting garbage collected on reload. Although, I am not very experienced in using the memory dev tools.

If you are concerned I could pull out wrap up codes all together or do another refactor.

examples/interactionsDemo.html
justinaray: 👍

examples/interactionsDemo.html
justinaray: Yeah, I was worried about both the click listener and the promise in terms of leaking.  I did a little research and it doesn't seem like listeners would hold a reference implicitly.  And, I wasn't sure if the memory would be free'd once the promise resolved if the dom was torn down before.  That was my major concern.

If it doesn't appear to be leaking, I'm fine with that as it is an example/testing tool after all.

justinaray: image

justinaray: Yeah, memory seems to be released on GC.  And listeners are growing but then shrinking back down and stable at 61 after quite a few runs.  With only 16 rows, that seems a bit high for a listener count, but it could be frameworks, etc.  Also, since it's not growing, that's good.

I wasn't gentle.  Clicked multiple times while the table was loading, so I thik we're ok from a leak standpoint.

src/index.js
```diff
@@ -148,6 +149,20 @@ class ClientApp {
...
+         * The ConversationsApi instance.
+         *
+         * @type {module:modules/conversations~ConversationsApi}
+         *
+         * @example
+         * let clientApp = new ClientApp({
+         *   pcEnvironmentQueryParam: 'pcEnvironment'
+         * });
+         *
+         * clientApp.conversationsApi.someMethod(...);
```
justinaray: Just `conversations`.  Will fix.

PR #54, opened 2019-09-04
Lifecycle improvement

feelbergood:


examples/lifecycleDemo.html
```diff
@@ -136,12 +137,13 @@ <h1>Lifecycle Events Log:</h1>
...
                         myClientApp.alerting.showToastPopup(
                             lifecycleStatusMessageTitle,
-                            'Bootstrap Complete (500ms delay)', {
-                                id: lifecycleStatusMessageId,
-                                type: 'success'
+                            `Bootstrap Complete (500ms delay):lifecycleStatusMessageId-bootstrap-${count}`, {
+                                id: `lifecycleStatusMessageId-bootstrap-${count++}`,
+                                type: 'success',
+                                showCloseButton: true,
+                                timeout: 0
```
justinaray: I wonder if we shouldn't add a querystring parameter to the url so that these "sticky" notifications aren't annoying to users using the examples in other contexts.  I personally use this example for manual regression testing.

Maybe add something like requireManualToastDismissal=true?

examples/lifecycleDemo.html
```diff
@@ -136,12 +137,13 @@ <h1>Lifecycle Events Log:</h1>
...
                     window.setTimeout(() => {
                         myClientApp.lifecycle.bootstrapped();
-
                         myClientApp.alerting.showToastPopup(
                             lifecycleStatusMessageTitle,
-                            'Bootstrap Complete (500ms delay)', {
-                                id: lifecycleStatusMessageId,
-                                type: 'success'
+                            `Bootstrap Complete (500ms delay):lifecycleStatusMessageId-bootstrap-${count}`, {
+                                id: `lifecycleStatusMessageId-bootstrap-${count++}`,
```
justinaray: looked at the spec and it appears that function arguments are evaluated ltr so these counts should be sync'd.  Has this been the case in your testing?  Even so, I wonder if we should just increment beforehand and not rely on order of function argument evaluation.

examples/lifecycleDemo.html
```diff
@@ -183,15 +271,27 @@ <h1>Lifecycle Events Log:</h1>
...
-                            'App Stopped (500ms delay)', {
-                                id: lifecycleStatusMessageId,
-                                type: 'error',
-                                showCloseButton: true
-                            }
-                        );
+                        if (requireManualToastDismissal) {
+                            myClientApp.alerting.showToastPopup(
+                                lifecycleStatusMessageTitle,
+                                `App Stopped (500ms delay):lifecycleStatusMessageId-stop-${count}`, {
```
justinaray: This is a different format than the rest of the app messages.  Still contains the `(500ms delay)` string.

examples/lifecycleDemo.html
```diff
@@ -115,9 +115,61 @@ <h1>Lifecycle Events Log:</h1>
...
+                    });
+                }
+
+                return result;
+            }
+
+            function getRequireManualToastDismissal() {
+                let result = null;
+
+                if (window.location.hash && window.location.hash.indexOf('access_token') >= 0) {
```
justinaray: This example doesn't require PC Auth, so no need for this block.  You can just check the search string directly as you're doing on 160.

examples/lifecycleDemo.html
justinaray: whoops.  nevermind.  Was comparing with focus/blur.  Disregard this comment.

examples/lifecycleDemo.html
```diff
@@ -136,40 +188,76 @@ <h1>Lifecycle Events Log:</h1>
...
-                            lifecycleStatusMessageTitle,
-                            'Bootstrap Complete (500ms delay)', {
-                                id: lifecycleStatusMessageId,
-                                type: 'success'
-                            }
-                        );
+                        if (requireManualToastDismissal) {
+                            myClientApp.alerting.showToastPopup(
+                                lifecycleStatusMessageTitle,
+                                `Bootstrap Complete (500ms delay):lifecycleStatusMessageId-bootstrap-${count}`, {
```
justinaray: Did you mean to use lifecycleStatusMessageId as a variable or string here?

You could pull this all out to a utility method passing desc base, type, and requireManualToastDismissal.

```js
_showLifecycleToast('Bootstrap Complete (500ms delay)', 'success', requireManualToastDismissal);

function _showLifecycleToast(msgBody, msgType, requireManualToastDismissal, msgTitle) {
let finalMsgTitle = msgTitle || lifecycleStatusMessageTitle;
// use the args here to build the correct msgBody and options based on requireManualToastDismissal
}
```

examples/lifecycleDemo.html
```diff
@@ -136,40 +188,76 @@ <h1>Lifecycle Events Log:</h1>
...
+                                    showCloseButton: true,
+                                    timeout: 0
+                                }
+                            );
+                            count++;
+                        } else {
+                            myClientApp.alerting.showToastPopup(
+                                lifecycleStatusMessageTitle,
+                                `Bootstrap Complete (500ms delay)`, {
+                                    id: `lifecycleStatusMessageId`,
```
justinaray: ^ this uses the string and not the variable, so the result is an id with the name of the variable vs 'lifecycleDemo-statusMsg'

examples/lifecycleDemo.html
```diff
@@ -115,9 +115,51 @@ <h1>Lifecycle Events Log:</h1>
...
+                    });
+                }
+
+                return result;
+            }
+
+            function getRequireManualToastDismissal() {
+                let result = null;
+
+                if (!result && window.location.search) {
```
justinaray: You dont need the !result check here anymore since the oauth params were removed.
You could moce the 'true' value check here and just return a boolean from this method

examples/lifecycleDemo.html
```diff
@@ -136,40 +178,76 @@ <h1>Lifecycle Events Log:</h1>
...
+                                    showCloseButton: true,
+                                    timeout: 0
+                                }
+                            );
+                            count++;
+                        } else {
+                            myClientApp.alerting.showToastPopup(
+                                lifecycleStatusMessageTitle,
+                                `Bootstrap Complete (500ms delay)`, {
+                                    id: `${lifecycleStatusMessageId}`,
```
justinaray: you could just do `id: lifecycleStatusMessageId` here.

examples/lifecycleDemo.html
feelbergood: I'm not sure if I totally understand this...is it something like
```if (window.location.search) {
                    return extractParams(window.location.search.substring(1)).requireManualToastDismissal;
                }
                return null;```?

Comment (justinaray) [reply examples/lifecycleDemo.html]: Yeah, that was 2 comments in 1.  Sorry.

The first comment was to remove the !result check as it was a hold-over from the previous code that first checked for the param in the hash after an authentication attempt.

The second comment is that you could move the parsing of the param value into this utility as well.  So, rather than returning the value of the parameter, you could return a boolean from this method indicating if the toasts need to be manually dismissed.

Comment (feelbergood) [reply examples/lifecycleDemo.html]: I assume we will add to the url something like requireManualToastDismissal=true/false. So I change the code to:
```
let result = null;
if (window.location.search) {
   result = extractParams(window.location.search.substring(1)).requireManualToastDismissal;
   if (result === 'true') {
      return true;
   }
}
return false;
```
I'm wondering if this is correct?

File: examples/lifecycleDemo.html [hunk 9]
```diff
@@ -115,9 +115,52 @@ <h1>Lifecycle Events Log:</h1>
...
+                        return true;
+                    }
+                }
+                return false;
+            }
+
             jQuery(function () {
                 let myClientApp = null;
 
+                var requireManualToastDismissal = getRequireManualToastDismissal() === 'true' ? true : false;
```
justinaray: This line won't work now.  You should just use the return value of the function now that it directly returns the boolean.

PR #55, opened 2019-09-06
Add info on required lifecycle hooks to docs

justinaray:
Add information about lifecycle hook config required for the various utils.


justinaray:
@L-L-the-noob would these doc changes have helped with the confusion you encountered?

src/modules/lifecycle.js
```diff
@@ -20,8 +20,32 @@ const STOP_HOOK_FILTER = HOOK_NAME_FILTER.bind(undefined, 'stop');
...
+ *       "stop": <boolean>,
+ *       "blur": <boolean>,
+ *       "focus": <boolean>,
+ *       "bootstrap": <boolean>
+ *     }
+ *   }
+ * }
+ * ```
+ *
+ * See [Advanced Application Concepts - Lifecycle Events](https://developer.mypurecloud.com/api/client-apps/advanced.html#lifecycle_events) for more details.
```
PrinceNebulon: Use the path only so it works in all regions: `[Advanced Application Concepts - Lifecycle Events](/api/client-apps/advanced.html#lifecycle_events)`. Same for the rest of the URLs.

PrinceNebulon: I just realized this is a javascript file, not markdown docs for the dev center. A full URL may be more appropriate here in that case.

justinaray: It gets piped through jsdoc2md so it will end up as markdown after the build.

Relative links should work at that point, right?  If so, thanks for the suggestion.  Much better.

justinaray: Updated ... thanks for the catch

PR #58, opened 2019-12-04
Feature/show external profile

griffind:


examples/externalProfileDemo.html
```diff
@@ -0,0 +1,337 @@
...
+    -->
+    <head>
+        <meta charset="utf-8">
+        <title>PureCloud client-app-sdk External Contacts Demo</title>
+
+        <link rel="stylesheet" href="https://maxcdn.bootstrapcdn.com/bootstrap/3.3.5/css/bootstrap.min.css" integrity="sha512-dTfge/zgoMYpP7QbHy4gWMEGsbsdZeCXz7irItjcC3sPUFtf0kuFbDz/ixG7ArTxmDjLXDmezHubeNikyKGVyQ==" crossorigin="anonymous">
+        <link rel="stylesheet" href="https://maxcdn.bootstrapcdn.com/bootstrap/3.3.5/css/bootstrap-theme.min.css" integrity="sha384-aUGj/X2zp5rLCbBxumKTCw2Z50WgIr1vs/PFN4praOTvYXWlVyh2UtNUU0KAUhAX" crossorigin="anonymous">
+
+        <script src="https://sdk-cdn.mypurecloud.com/javascript/52.1.1/purecloud-platform-client-v2.min.js"></script>
+        <script src="https://sdk-cdn.mypurecloud.com/client-apps/1.3.0/purecloud-client-app-sdk-43593954923a1eb6031c0828e0d4d5d1.min.js"></script>
```
justinaray: As part of another PR, I added support for multiple files.  So, if you want, you can move your js and css into separate files and get proper syntax highlighting.

It still doesn't support a build, but separate files is a step in the right direction.

examples/externalProfileDemo.html
```diff
@@ -0,0 +1,337 @@
...
+                                let entities = data.entities;
+                                // Build table with data from api call
+                                for (let i = 0; i < entities.length; i++) {
+                                    document.getElementById('tableBody').insertAdjacentHTML('beforeend',
+                                        `<tr>
+                                            <td>
+                                                <a id="entity-${i}">${entities[i].lastName}, ${entities[i].firstName}</a>
+                                            </td>
+                                            <td>
+                                                <a id="org-${i}">${entities[i].externalOrganization.id}</a>
```
justinaray: This is failing in my org.  Seems like I have a contact without an org.  So, you'll probably need to check for that before adding this table cell.

examples/externalProfileDemo.html
```diff
@@ -0,0 +1,337 @@
...
+                    document.querySelector('.progress-bar').className += ' progress-bar-danger';
+                    updateProgressBar(100);
+                }
+
+                /**
+                 * Parse the permissions array and check whether or not the match the specified required ones.
+                 *
+                 * @returns A boolean indicating if the user possesses the required permissions.
+                 */
+                function checkPermission(permissions, permissionValue) {
```
justinaray: I need to see if my new build process will allow for us to pull stuff like this and query string parsing out into a utils folder at this same level.  No need to do it for this PR, but it just reminded me.

src/index.js
```diff
@@ -19,6 +19,7 @@ import CoreUiApi from './modules/ui';
 import UsersApi from './modules/users';
 import ConversationsApi from './modules/conversations';
 import MyConversationsApi from './modules/myConversations';
+import ContactsApi from './modules/contacts.js';
```
justinaray: Would you mind changing from `contacts` to `externalContacts` for consistency with the public api?  Sorry!  I know this will pretty much touch everything in the PR (filename, module name, examples, test).  Hope it's not too much of a PITA.

Looks like I'm using camel case for most things

examples/externalProfileDemo.html
```diff
@@ -0,0 +1,337 @@
...
+                 * Determine the embedding PureCloud environment seeded on the query string or
+                 * being returned through the OAuth2 Implicit grant state hash param.
+                 *
+                 * @returns A string indicating the embedding PC env (e.g. mypurecloud.com, mypurecloud.jp); otherwise, null.
+                 */
+                function getEmbeddingPCEnv() {
+                    let result = null;
+
+                    if (window.location.hash && window.location.hash.indexOf('access_token') >= 0) {
+                        let oauthParams = extractParams(window.location.hash.substring(1));
```
justinaray: extractParams is actually pulled in from a utils.js in the other example that uses it.  You'll either want to copy that utils script and pull it in or we can look to make it a common set of utils.  Again, need to check the build to see if that's supported

examples/externalProfileDemo.html
```diff
@@ -0,0 +1,337 @@
+<!DOCTYPE html>
```
justinaray: maybe we should change this to externalContactsDemo.html?

examples/externalContactsDemo/script.js
```diff
@@ -0,0 +1,269 @@
...
+                                </td>
+                            </tr>`
+                        );
+                        // Attach listeners to entities to call sdk methods
+                        document.getElementById(`entity-${i}`).addEventListener('click', function () {
+                            clientApp.externalContacts.showExternalContactProfile(entities[i].id);
+                        });
+                        if (hasOrg) {
+                            document.getElementById(`org-${i}`).addEventListener('click', function () {
+                                clientApp.externalContacts.showExternalOrganizationProfile(entities[i].externalOrganization.id);
```
griffind: `clientApp.externalContacts` is undefined here as the module seems to no longer be in the clientApp closure since I separated this example out into its own directory. Not sure why that would be the case. It worked before and the test for Modules passes (`'should have instances of all the modules'`) in `ClientAppSpec.js` 🤷

examples/externalContactsDemo/index.html
```diff
@@ -0,0 +1,78 @@
...
+
+    <link rel="stylesheet" href="style.css">
+    <link rel="stylesheet" href="https://maxcdn.bootstrapcdn.com/bootstrap/3.3.5/css/bootstrap.min.css"
+        integrity="sha512-dTfge/zgoMYpP7QbHy4gWMEGsbsdZeCXz7irItjcC3sPUFtf0kuFbDz/ixG7ArTxmDjLXDmezHubeNikyKGVyQ=="
+        crossorigin="anonymous">
+    <link rel="stylesheet" href="https://maxcdn.bootstrapcdn.com/bootstrap/3.3.5/css/bootstrap-theme.min.css"
+        integrity="sha384-aUGj/X2zp5rLCbBxumKTCw2Z50WgIr1vs/PFN4praOTvYXWlVyh2UtNUU0KAUhAX" crossorigin="anonymous">
+
+    <script src="https://sdk-cdn.mypurecloud.com/javascript/52.1.1/purecloud-platform-client-v2.min.js"></script>
+    <script src="https://sdk-cdn.mypurecloud.com/client-apps/1.3.0/purecloud-client-app-sdk-43593954923a1eb6031c0828e0d4d5d1.min.js"></script>
```
griffind: This is where my confusion stemmed. This is pulling in `1.3.0` which im guessing will have to be changed to `1.4.0` which is when this end point will be active. I think this works now because it is using this branch version of the sdk?

justinaray: yup, it's using the branch's local version.  Like we discussed, we'll just bump this and all examples when 1.4.0 lands.

src/modules/externalContacts.js
```diff
@@ -0,0 +1,51 @@
+/**
+ * Utilities for interacting with PureCloud conversations
```
justinaray: `with PureCloud External Contacts`.  Actually, with the name change inbound, maybe change this to just `with Extrernal Contacts`

src/modules/externalContacts.js
```diff
@@ -0,0 +1,51 @@
...
+ *
+ * @module modules/externalContacts
+ *
+ * @since 1.4.0
+ */
+
+import BaseApi from './base';
+
+/**
+ * Utilities for interacting with PureCloud conversations
```
justinaray: here as well

src/ClientAppSpec.js
```diff
@@ -17,7 +17,7 @@ export default describe('ClientApp', () => {
 
     it('should have instances of all the modules', () => {
         let myClientApp = new ClientApp();
-        ['alerting', 'lifecycle', 'coreUi', 'users'].forEach(currModuleName => {
+        ['alerting', 'lifecycle', 'coreUi', 'users', 'myConversations', 'externalContacts'].forEach(currModuleName => {
```
justinaray: Nice catch!  Thanks for adding myConversations.

examples/externalContactsDemo/script.js
```diff
@@ -0,0 +1,269 @@
...
+    /*
+    * Note: To use this app in your own org, you will need to create your own OAuth2 Client(s)
+    * in your PureCloud org.  After creating the Implicit grant client, map the client id(s) to
+    * the specified region key(s) in the object below, deploy the page, and configure an app to point to that URL.
+    */
+    let pcOAuthClientIds = { 'mypurecloud.com': 'implicit-oauth-client-id-here' };
+    let clientId = pcOAuthClientIds[pcEnvironment];
+    if (!clientId) {
+        setErrorState(
+            pcEnvironment + ': Unknown/Unsupported PureCloud Environment'
```
justinaray: Since this is user data, I'll update the setErrorState method to use innerText.

examples/externalContactsDemo/script.js
```diff
@@ -0,0 +1,269 @@
...
+        let apiInstance = new platformClient.ExternalContactsApi();
+        let opts = {
+            pageSize: 6,
+            pageNumber: 1
+        };
+        return apiInstance.getExternalcontactsContacts(opts)
+            .then(data => {
+                return data;
+            }).catch(err => {
+                setErrorState(err);
```
justinaray: Catching here actually resolves the promise with undefined which will then throw a reference error in the then handler above.  I would recommend you dont catch here and deal with it above.

examples/externalContactsDemo/script.js
```diff
@@ -0,0 +1,269 @@
...
+                            clientApp.externalContacts.showExternalContactProfile(entities[i].id);
+                        });
+                        if (hasOrg) {
+                            document.getElementById(`org-${i}`).addEventListener('click', function () {
+                                clientApp.externalContacts.showExternalOrganizationProfile(entities[i].externalOrganization.id);
+                            });
+                        }
+                    }
+                });
+                // Hide progress bar after auth is done
```
justinaray: Would this be cleaner in a finally at the bottom of this setup promise chain?  Not sure of all the edge cases, so feel free to not do this if it doesn't make sense.

examples/externalContactsDemo/script.js
```diff
@@ -0,0 +1,269 @@
...
+                    let entities = data.entities;
+                    // Build table with data from api call
+                    for (let i = 0; i < entities.length; i++) {
+                        let hasOrg = !!entities[i].externalOrganization;
+                        document.getElementById('tableBody').insertAdjacentHTML('beforeend',
+                            `<tr>
+                                <td>
+                                    <a id="entity-${i}">${entities[i].lastName}, ${entities[i].firstName}</a>
+                                </td>
+                                <td>
```
justinaray: Do you need ot check the org view permission before adding the org links?

examples/externalContactsDemo/script.js
```diff
@@ -0,0 +1,269 @@
...
+            if (checkPermission(permissions, 'externalContacts:contact:view')) {
+                requestExternalContacts().then(data => {
+                    let entities = data.entities;
+                    // Build table with data from api call
+                    for (let i = 0; i < entities.length; i++) {
+                        let hasOrg = !!entities[i].externalOrganization;
+                        document.getElementById('tableBody').insertAdjacentHTML('beforeend',
+                            `<tr>
+                                <td>
+                                    <a id="entity-${i}">${entities[i].lastName}, ${entities[i].firstName}</a>
```
justinaray: I wonder if you could create the nodes first and then come back through with innerText to insert this user data so it can't be XSS?  Sorry, I know this is a PITA and just a demo app, but prob worth doing.

PR #67, opened 2020-05-14
ATEAM-253: Fix broken docs links

jakejrichards:
It looks like the reason that these links are broken and the others aren't is because we have `ClientApp.md` hardcoded in the `doc/index.md` file and we just need to change that to `ClientApp.html`. Also, while I was in the docs I also noticed that we were missing the return type for the `ClientApp.about` method, so I went ahead and added that bonus change here.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -71,6 +71,11 @@ if (SUPPORTED_DOC_OUTPUT_FORMATS.indexOf(docMdOutputFormat) < 0) {
     process.exit(1);
 }
 
+let relativeLinkExtension = '.md';
+if (docMdOutputFormat === PC_DEV_CENTER_FORMAT) {
+    relativeLinkExtension = '.html';
+}
+
 var build = function () {
     return Promise.all([buildStdDists(), buildProdBrowserDist()]);
 };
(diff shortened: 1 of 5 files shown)


jakejrichards:
So we've decided to bring in some regex transformation to avoid hardcoding the `.html` extension in the `index.md` file. Now, all of the relative links in the docs will have the same specified extension. This also adds support for extensionless links (when/if the dev center allows this). Extensionless links also work well when navigating within markdown, so we could think about dropping `.md` extension support and just use either `.html` or extensionless.

gulpfile.js
```diff
@@ -246,6 +259,13 @@ var buildDoc = function () {
     });
 };
 
+function transformRelativeLinks(buffer, ext) {
+    const RELATIVE_LINK_REPLACE_REGEXP = /\(\.\/(.+)\..+\)/gm;
```
justinaray, line 263: I wonder if we should tweak this slightly to `/\(\.\/([^)]+)\.[^)]+\)/gm` so that we can handle two instances of links in a line .. in case it comes up.

`[Foo](./foo.md) [Bar](./bar.md)`

gulpfile.js
```diff
@@ -71,6 +71,17 @@ if (SUPPORTED_DOC_OUTPUT_FORMATS.indexOf(docMdOutputFormat) < 0) {
     process.exit(1);
 }
 
+let relativeLinkExtension = '.html';
```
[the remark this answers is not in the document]
jakejrichards: Yeah, that seems reasonable. Basically we would just have `.md` in all cases except `PC_DEV_CENTER_FORMAT`.

gulpfile.js
jakejrichards: Yeah, that makes sense to me, this is some tier 1 regex.

PR #68, opened 2020-06-04
ATEAM-266: Move docs script out of gulpfile

jakejrichards:
The first step toward removing gulp and cleaning up the build process is to move the individual scripts out of the gulpfile. This PR just moves the build docs script to its own individual script file that can still be run via "npm run docs".

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -4,11 +4,8 @@
 var pkg = require('./package.json');
 var path = require('path');
 var fs = require('fs');
-let fsThen = require('fs-then-native');
-var mkdirp = require('mkdirp');
 
 var gulp = require('gulp');
-var gutil = require('gulp-util');
 var del = require('del');
 var replace = require('gulp-replace');
 var runSequence = require('run-sequence');
@@ -22,6 +19,8 @@ var rollupPluginUglify = require('rollup-plugin-uglify');
 
 // Development Server Utils
 var browserSync = require('browser-sync');
+
+const { buildDocs } = require('./scripts/build-docs');
 
 const DEST_DIR = 'dist';
 const GLOBAL_LIBRARY_NAME = 'purecloud.apps.ClientApp';
@@ -51,30 +50,6 @@ const DEFAULT_ROLLUP_CONFIG = {
     ],
     cache: false
 };
-
-let flagDocsAsPreview = true;
-if (process.env.FLAG_DOCS_AS_PREVIEW) {
-    ['false', 'f', '0'].forEach(currFalseValue => {
-        if (process.env.FLAG_DOCS_AS_PREVIEW === currFalseValue) {
-            flagDocsAsPreview = false;
-        }
-    });
-}
-
-const GITHUB_FORMAT = 'github';
-const KRAMDOWN_FORMAT = 'kramdown';
-const PC_DEV_CENTER_FORMAT = 'purecloudDevCenter';
-const SUPPORTED_DOC_OUTPUT_FORMATS = [GITHUB_FORMAT, KRAMDOWN_FORMAT, PC_DEV_CENTER_FORMAT];
-let docMdOutputFormat = process.env.DOC_MD_OUTPUT_FORMAT || PC_DEV_CENTER_FORMAT;
-if (SUPPORTED_DOC_OUTPUT_FORMATS.indexOf(docMdOutputFormat) < 0) {
-    console.error(`Unknown MD Output Format Specified: '${docMdOutputFormat}'`);
-    process.exit(1);
-}
-
-let relativeLinkExtension = '.md';
-if (docMdOutputFormat === PC_DEV_CENTER_FORMAT) {
-    relativeLinkExtension = '.html';
-}
 
 var build = function () {
     return Promise.all([buildStdDists(), buildProdBrowserDist()]);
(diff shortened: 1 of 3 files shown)


scripts/build-docs.js
```diff
@@ -0,0 +1,158 @@
...
+        if (process.env.FLAG_DOCS_AS_PREVIEW === currFalseValue) {
+            flagDocsAsPreview = false;
+        }
+    });
+}
+
+// If invoked via "npm run docs" or "node scripts/build-docs.js" run immediately
+const calledFromCli = !module.parent;
+if (calledFromCli) {
+    buildDocs();
```
justinaray, line 36: Thoughts on adding a catch handler here, log to the console, and then process.exit(1).  This will decouple the impl of buildDocs from it's invoking context.

scripts/build-docs.js
```diff
@@ -0,0 +1,158 @@
...
+const KRAMDOWN_FORMAT = 'kramdown';
+const PC_DEV_CENTER_FORMAT = 'purecloudDevCenter';
+const SUPPORTED_DOC_OUTPUT_FORMATS = [GITHUB_FORMAT, KRAMDOWN_FORMAT, PC_DEV_CENTER_FORMAT];
+let docMdOutputFormat = process.env.DOC_MD_OUTPUT_FORMAT || PC_DEV_CENTER_FORMAT;
+if (SUPPORTED_DOC_OUTPUT_FORMATS.indexOf(docMdOutputFormat) < 0) {
+    console.error(`Unknown MD Output Format Specified: '${docMdOutputFormat}'`);
+    process.exit(1);
+}
+
+let relativeLinkExtension = '.md';
```
justinaray, line 19: nit, this should be linkExtension since it's global in this context.

scripts/build-docs.js
```diff
@@ -0,0 +1,158 @@
...
+                        return fsThen.writeFile(path.join(docDestDirPath, `${className}.md`), output);
+                    })
+                );
+            }
+        });
+
+        return Promise.all(docPromises);
+    }).catch(reason => {
+        let errMsg = 'Documentation generation failed';
+        console.error(errMsg, reason);
```
justinaray, line 153: i think we could reject here with an object.  maybe something like {msg, reason}.  This lets the caller deal with how to handle ... logging (and which logger to use) and/or process.exit.

This is related to the cli handler comment above.

package.json
```diff
@@ -74,7 +74,6 @@
     "gh-pages": "^0.11.0",
     "gulp": "^3.9.1",
     "gulp-replace": "^0.5.4",
-    "gulp-util": "^3.0.8",
```
justinaray, line 77: Nice!  first dep gone!

Can you remove this with yarn so we clean up the lockfile as well?  Hopefully you can remove without updating the rest of the deps as that would be out of scope here.

jakejrichards: Yeah, I was running into issues with yarn. Which led me down this route: https://github.com/yarnpkg/yarn/issues/5962. I think our only solution is to remove the lockfile and reinstall (which unfortunately updates the rest of the deps). Maybe what I'll do is open a PR to be merged before this that fixes our lockfile.

scripts/build-docs.js
jakejrichards: Yeah, the "relative" here applied to the fact that they were relative links. But then I see that I named it `linkExtension` in the helper 🙃

scripts/build-docs.js
jakejrichards: I think it makes sense then to not catch the error at all, and just leave it up to the caller to handle the error, right?

package.json
justinaray: yarn.lock isn't being updated as this is still a transitive dependency from some other deps.  Still, it's days are numbered!

justinaray:
Thanks for tackling this!  very cool!

I diffed the dist from master and this branch and they are identical ... accept for the fixed links and return type of course.

scripts/build-docs.js
justinaray: yeah, i think that makes sense here since we'll mostly just catch and log it anyway.  if it was an api, maybe we'd come up with a few types of errors (enum) and throw that and the nitty gritty deets.

mostly this will just get rid of a log statement here in case it is handled by the caller and the log would be a red herring.

scripts/build-docs.js
```diff
@@ -0,0 +1,155 @@
+/* eslint-env node */
```
justinaray, line 1: Thoughts on adding a shebang and making this script executable?  Of course, we could do this later when removing gulp or not do it at all if we want to ask users to call through npm run which makes sense as well.

gulpfile.js
```diff
@@ -293,16 +149,21 @@ var buildExamples = function (destPath = DEST_DIR, sdkUrl = null) {
...
-    runSequence('clean', ['build', 'doc'], done);
+    runSequence('clean', ['build', 'docs'], done);
 });
 
 gulp.task('clean', function () {
     return del([DEST_DIR]);
 });
 
-gulp.task('build', build);
+gulp.task('docs', () => {
```
justinaray, line 159: This name change would change contributing.md.  But, we're dumping gulp soon enough and you've got it removed in a later commit, so no need to make that change here.

scripts/build-docs.js
jakejrichards: I just googled this and didn't even know it was a thing! I figured the only way to do it was an npm script. We can always come back and add it later if it's something we feel we need. I'll note it.

PR #69, opened 2020-06-04
ATEAM-266: Move build examples script out of gulpfile

jakejrichards:
This PR moves the `buildExamples` function out of gulpfile and into its own script file that can be invoked via `npm run examples`. I also needed to switch to a vanilla js implementation, which isn't too bad. I pulled in `fs-extra` for the added functionality to make it easier to implement, and since I did that I went ahead and dropped `fs-then-native` since `fs-extra` has all the same features and more.


scripts/build-examples.js
```diff
@@ -0,0 +1,37 @@
...
+        const res = path.resolve(dir, subdir);
+        return (await fs.stat(res)).isDirectory() ? getFiles(res) : res;
+    }));
+    return files.reduce((a, f) => a.concat(f), []);
+}
+
+async function buildExamples(destPath = DEST_DIR, sdkUrl = null) {
+    await fs.copy('examples', destPath);
+    const files = await getFiles(destPath);
+    for (let file of files) {
```
justinaray, line 27: I dont think we're guaranteed of a clean dist directory here.  there might be other build artifacts that could match and get corrupted.  I wonder if we should use a temp directory to mutate and then copy into dest_dir?  Or, we could look into streaming options?

package.json
```diff
@@ -20,6 +20,7 @@
     "lint": "eslint gulpfile.js src/*.js src/**/*.js",
     "build": "gulp",
     "docs": "node scripts/build-docs.js",
+    "examples": "node scripts/build-examples.js",
```
justinaray, line 23: Do we need a separate build script for the examples?  Might just be part of `start` `start:watch`.  I don't have a strong opinion here.  Just raising it as a thought.

gulpfile.js
```diff
@@ -173,8 +167,8 @@ gulp.task('watch', function () {
 });
 
 gulp.task('serve', ['clean'], function () {
-    let buildSdkForDevServer = buildStdDists.bind(this, path.resolve(DEST_DIR, 'vendor'));
-    let buildExamplesForDevServer = buildExamples.bind(this, DEST_DIR, `vendor/${BROWSER_FILENAME}`);
+    let buildSdkForDevServer = () => buildStdDists(path.resolve(DEST_DIR, 'vendor'));
```
justinaray, line 170: No more bind for scope, yay!

scripts/build-examples.js
jakejrichards: Yeah, that's a fair point. I'll look into this!

package.json
jakejrichards: Yeah... reflecting on it, building the examples without first running the start script wouldn't make much sense.

justinaray: I think we should remove this script then since you removed the cli support from the script.

jakejrichards: Oh, that's strange I could've sworn I removed it with my most recent commit. I'll fix it up, good catch!

PR #70, opened 2020-06-04
ATEAM-266: Remove gulp entirely and move into a rollup config file

jakejrichards:
So this is the meat of the move away from gulp. I got all of the build config moved into a `rollup.config.js` file and bumped our rollup version to latest. The version bump allowed me to drop `hasha` and use a new output option `entryFileNames` which will automatically inject a content hash into a template string `[hash]` - so that's nice. I also had to drop the rollup uglify plugin in favor of terser since uglify is no longer maintained nor compatible with rollup 7.x. Future improvements are to:
- upgrade the rest of the rollup plugins
- pull in the `selfsigned` npm package to auto-generate SSL cert/key for us 
- add live reloading to the dev server (automatic page refresh when src or example code changed)


package.json
```diff
@@ -51,7 +52,7 @@
   },
   "license": "MIT",
   "engines": {
-    "node": ">= 6.9.4"
+    "node": ">= 12.18.0"
```
justinaray, line 55: # :tada:

package.json
```diff
@@ -17,15 +17,16 @@
     "examples": "examples"
   },
   "scripts": {
-    "lint": "eslint gulpfile.js src/*.js src/**/*.js",
-    "build": "gulp",
+    "lint": "eslint src/*.js src/**/*.js",
+    "start": "DEV_SERVER=true node scripts/dev-server",
+    "build": "./scripts/build.sh",
+    "rollup": "rollup --config",
```
justinaray, line 23: Do we need a rollup script?  Haven't yet seen if you're using it below.  Def fine if so.

justinaray: and ... i see the usage.  nothing to see here ;)

rollup.config.js
```diff
@@ -0,0 +1,105 @@
...
+        babel({
+            exclude: 'node_modules/**'
+        })
+    ]
+};
+
+if (process.env.DEV_SERVER) {
+    baseRollupConfig.plugins.push(serve({
+        host: 'localhost',
+        port: 8443,
```
justinaray, line 45: future enhancement could be to parameterize this, but def good with this for now.

scripts/dev-server.js
```diff
@@ -0,0 +1,35 @@
...
+const loadConfigFile = require('rollup/dist/loadConfigFile');
+const path = require('path');
+const rollup = require('rollup');
+const chokidar = require('chokidar');
+
+const { buildExamples } = require('./build-examples');
+
+const DEST_DIR = 'dist';
+const BROWSER_FILENAME = `${pkg.name}.js`;
+
```
justinaray, line 14: Wonder if we should flush the dist dir before starting this?

scripts/dev-server.js
```diff
@@ -0,0 +1,35 @@
+/* eslint-env node */
```
justinaray, line 1: I played with the rollup plugin api tonight and we may be able to do this without this script.  Not ready to PR to this, so I'll just show it to you tomorrow and we can discuss.

This works fine by the way of course, but, since we're relying on the rollup serve plugin anyway, keeping it in rollup would make sense.

justinaray:
:fire:

So much cleanup and good stuff in here.  Nice!

scripts/dev-server.js
jakejrichards: Now that I'm thinking about it, I think it makes more sense to serve the dev build + examples from a different folder than `dist` which contains docs and the other build files.

Either way, I think flushing the output dir is a good call though.

jakejrichards:
Went ahead and pulled in the live reload plugin with the [latest commit](https://github.com/MyPureCloud/client-app-sdk/pull/70/commits/531427b33bb7b54a266f2d7c8aa8a5a22e53991d) and also realized that `jsdoc-to-markdown` needed to be upgraded for node 12 support. I diffed the old docs vs new docs and no changes, so we should be good 👍

dev.rollup.config.js
```diff
@@ -0,0 +1,66 @@
...
+    fs.outputFileSync(
+        path.join(tmpDestPath, relativeFilePath.replace('examples/', '')),
+        transformExampleSdkUrl(fs.readFileSync(relativeFilePath, 'utf8'))
+    );
+};
+
+// Build all examples initially
+glob.sync('examples/**/*', { nodir: true }).forEach(buildExample);
+
+export default Object.assign({}, umdConfig, {
```
justinaray, line 38: This is :fire: @jakejrichards!  Turned out great, works perfectly.  Tested watch/live reload for both sdk and examples.

dev.rollup.config.js
```diff
@@ -0,0 +1,66 @@
...
+import glob from 'glob';
+import serve from 'rollup-plugin-serve';
+import livereload from 'rollup-plugin-livereload';
+import { umdConfig } from './rollup.config';
+
+const BROWSER_FILENAME = `${pkg.name}.js`;
+const tmpDestPath = fs.mkdtempSync(path.join(os.tmpdir(), `${pkg.name}-dev-server-`));
+
+const httpsConfig = {
+    key: fs.readFileSync('sslcert-dev/key.pem'),
```
justinaray, line 17: Should we wrap this with try/catch and either warn and fall back to http or fall if these files don't exist?  Not sure if any of these plugins strictly require https.  That would make the choice easy.

rollup.config.js
```diff
@@ -0,0 +1,93 @@
...
+*/`;
+
+const baseRollupConfig = {
+    input: './src/index.js',
+    cache: false,
+    plugins: [
+        commonjs(),
+        resolve(),
+        json(),
+        babel({
```
justinaray, line 34: I know this is copied from just what I had, but I'm trying to remember how the babel config worked way back when.  Not sure if it was pulling in the babel plugins automatically and if we're still getting them.  Looks like we have a few polyfills in there like object assign and external helpers.  Also, it references preset-latest which is deprecated.

For this lib, I figure we keep it pretty unopinionated and mostly let rollup deal with the modules and keep it pretty vanilla.  But, for the umd bundle for the cdn, maybe we need different babel config using presets?

Also, we can punt on this if OBE'd by follow on typescript work.

Let's chat a bit tomorrow about this.

justinaray: The diffs from current master and this branch are pretty close though, so, I think we're close.

rollup.config.js
```diff
@@ -0,0 +1,93 @@
...
+            entryFileNames: CJS_FILENAME,
+            format: 'cjs'
+        }
+    }),
+    buildRollupConfig({
+        output: {
+            banner: LEGAL_TEXT,
+            name: GLOBAL_LIBRARY_NAME,
+            dir: DEST_DIR,
+            entryFileNames: `${pkg.name}-[hash]${PROD_BROWSER_EXT}`,
```
justinaray, line 77: Is this using internal rollup hashing or rollup-plugin-hash?  I can see the plugin isn't used, but not sure if it's part of rollup proper now?  Looks like the default hash algorithm of the plugin is sha1.

Not that I mind, sha1 is obviously better, but, I noticed that the generated hash length is shorter than the previous md5.  Wondering how it's determining the necessary precision.

jakejrichards: Yeah, this is now using the template string that's built into rollup https://rollupjs.org/guide/en/#outputentryfilenames. They don't actually say in the docs what algorithm they're using, but I took a quick look at their source code and it seems to be sha256.

rollup.config.js
jakejrichards: This config will automatically pull in the presets/plugins from the `.babelrc` in the `src` folder and there shouldn't be any changes to the babel config in this PR. However, [in this PR](https://github.com/MyPureCloud/client-app-sdk/pull/71/files), I went ahead and upgraded babel for us and dropped preset-latest in favor of `@babel/preset-env` along with some other config changes. And then in the typescript PR there were also a few tweaks I had to make to the babel config. I'm happy to talk through this today!

dev.rollup.config.js
jakejrichards: Yeah, that's a good call. I don't believe that either of the plugins requires https so warning and falling back to HTTP would be a good approach here.

dev.rollup.config.js
```diff
@@ -0,0 +1,74 @@
...
+
+// Build all examples initially
+glob.sync('examples/**/*', { nodir: true }).forEach(buildExample);
+
+export default Object.assign({}, umdConfig, {
+    output: Object.assign({}, umdConfig.output, {
+        dir: tmpDestPath
+    }),
+    watch: {
+        clearScreen: false
```
jakejrichards, line 48: One thing I just added with the latest commit (to warn and fall back to HTTP) was to add this option here. Rollup was clearing the terminal on every build and all of our other logs were getting tossed. I'm open to leaving this as it was, we just wouldn't be able to have any other logs in our dev server.

justinaray: Seems fine to let the logs run to me.

PR #71, opened 2020-06-08
ATEAM-269: Update outdated rollup/babel deps

jakejrichards:
So the majority of the changes here were 1-1 straightforward replacements from outdated packages to new packages. A lot of the rollup plugins are now officially maintained by rollup and I got us upgraded to babel 7. The one actual change I did make was to replace `babel-plugin-external-helpers` with `@babel/plugin-transform-runtime`. I'm actually a bit confused about how the existing plugin was even working properly because as far as I could tell we weren't manually pulling in the helpers anywhere. Instead of figuring out how/where to do this, I opted to switch us over to runtime helpers, which is the recommended setting for libraries anyway (see here: https://github.com/rollup/plugins/tree/master/packages/babel#babelhelpers)


package.json
```diff
@@ -58,10 +58,14 @@
     "query-string": "^5.0.0"
```
justinaray, line 58: It sounds like we should add @babel/runtime as a dependency from the [docs](https://github.com/rollup/plugins/tree/master/packages/babel#babelhelpers).  Not sure if you noticed that or tried and had an issue?

jakejrichards: Yeah, really odd that I haven't run into any issues with this... I'll dive into it some more and see what's going on.

justinaray: Sounds good.  it's not clear on two fronts:

* is this a dep or devDep?  it's used at dev/build time, but it is also a runtime dep
* it says to "tell rollup to treat it as you external dependency when bundling for cjs & es"

justinaray: I've been digging into this this evening a bit as well.  The benefit of the runtime option is to both reduce the babel helper duplication and to provide a namespace in which to apply polyfills so you don't pollute the global namespace for downstream consumers.  Which sounds perfect, but ...

the rubs seem to be:
* only works in esm and cjs formats [ref](https://github.com/rollup/plugins/tree/master/packages/babel#injected-helpers)
* "The transformation plugin is typically used only in development" [ref](https://babeljs.io/docs/en/babel-plugin-transform-runtime#installation)

So, it seems we would potentially have a different set of babel plugins and rollup config for umd vs cjs/esm bundles.  And, potentially, another config for our dev server that's using the transform vs the external @babel/runtime dependency.

I tried adding @babel/runtime as a dep and specifying it as an external resource in rollup config but it didn't modify the output.  Presumably because we're already using inline everywhere as it's "hard-coded" in .babelrc.

So, tl;dr, seems like we've got more research to do to get all these packaged in the way we want them.  It looks like you can use getBabelOutputPlugin to control the babel plugins used (and presumably only have very common babel bits in babel.rc config?)  But, I don't love the idea of having every artifact use different babel config.  It may be necessary, but seems like a PITA.

Let's dig into this a bit tomorrow.

src/.babelrc
```diff
@@ -1,12 +1,9 @@
 {
     "presets": [
-      ["latest", {
-        "es2015": {
-            "modules": false
-        }
-      }]
+        ["@babel/preset-env", { "modules": false }]
```
justinaray, line 3: Looks like if no targets or browserslist is specified, then we'll get transpilation of all es2015+ code by default (except modules which is explicitly excluded).

This is definitely beneficial so we're not asking our consumers to transpile our code.  But, I wonder if we even need the present-env in this case since we won't be using it to target browsers with support?

Finally, rollup is recommending that libs not distribute non-transpiled code.  When I originally did the bundling, I took a different approach in that I expected an esm consumer to need to transpile.  But, with the hope of someday using esm directly, maybe this concept has changed.

Let's chat on this a bit tomorrow as well.

jakejrichards: So, I think we still need preset-env as without it babel won't transpile our code at all. I believe that babel's default behavior, with no presets/plugins, just leaves the code untouched. It seems that just using this preset, while it is heavy-handed when we don't have a browser list and by default compiling all es2015+ down to es5, is pretty reasonable. Perhaps we should consider what browsers to target that way we could be a bit more efficient here and compile less if we don't have to. I'm not sure what browsers we target as a company, but maybe we can get that list and just use that for the SDK as well. Let me know your thoughts on this.

rollup.config.js
```diff
@@ -24,49 +24,73 @@ const LEGAL_TEXT = `/*
...
+}, config);
 
-const buildRollupConfig = (config) => {
-    return Object.assign({}, baseRollupConfig, config);
-};
+const umdBabelOutputPlugin = getBabelOutputPlugin({
+    // Need to specify this for a umd output format
+    // https://github.com/rollup/plugins/tree/master/packages/babel#using-formats-other-than-es-modules-or-commonjs
+    allowAllFormats: true,
+    presets: ['@babel/env']
```
justinaray, line 44: Is `modules: false` not needed here?  I assumed we always wanted rollup to deal with imports, but maybe it has already happened by the time babel runs in the umd case?

jakejrichards: Oh man, do I have an answer for you on this one. So, I had to dig a bit for this one to be 100% positive (thank you garbage babel docs). The default behavior is `modules: "auto"`. Modules auto defaults to `false` if the current process is known to support ESM syntax. Relevant docs you'll only find in this PR: https://github.com/babel/babel/pull/8485. So there is actually a flag `supportsStaticESM` that can be specified when running babel directly. It just so happens that `@rollup/plugin-babel` supplies `supportsStaticESM: true` to babel in their implementation, [(see here)](https://github.com/rollup/plugins/blob/440dbc4f53c8ded33f2ad58f10ea647b30807435/packages/babel/src/index.js#L45) so we can safely drop `modules: false` from all of our builds. Thank you for coming to my TED talk.

I'm not opposed to just specifying `modules: false` everywhere to avoid confusion, what do you think?

justinaray: https://media.giphy.com/media/3ELtfmA4Apkju/giphy.gif

justinaray: haha .. thank you for digging that up.  I would be good with either always specifying and/or leaving a comment.  I feel like that might bite us in the future at some point.

jakejrichards: Yeah, I'm good with just specifying it everywhere as I don't want to have to go down that rabbit hole again 😆

jakejrichards:
@justinaray I rebased the typescript PR onto this one and dove into the babel config there as well. If you want to see the direction that I took and how this babel config plays out in the typescript world here's a [link to that file](https://github.com/MyPureCloud/client-app-sdk/blob/4640811602c23a555dd4998ef204a3c1397c73ae/rollup.config.js). Really, once the [`babelHelpers: runtime` bug](https://github.com/rollup/plugins/issues/475) is fixed, we'll be able to greatly simplify our babel config. But for now, this should get us where we need to be.

rollup.config.js
```diff
@@ -24,24 +24,28 @@ const LEGAL_TEXT = `/*
...
     ]
 };
 
-const buildRollupConfig = (config) => {
-    return Object.assign({}, baseRollupConfig, config);
-};
+const buildRollupConfig = (config) => Object.assign({}, baseRollupConfig, config);
+
+const umdBabelInputPlugin = getBabelInputPlugin({
+    babelHelpers: 'bundled',
```
justinaray, line 43: knowing what we now know about input and output rollup plugins, do you think we should have used transform-runtime in all cases and let rollup bundle up the dependencies based on the output format?  UMD would bundle it in vs cjs and esm would just leave the imports?

I'm not sure if this is possible, just curious now that we've learned more about all this.

justinaray: Even if that's possible but results in the helpers polluting the global namespace, I think this approach is better as we avoid pollution even if it's at the cost of our babel config complexity.

justinaray: Fiddled with this a bit tonight.  External seems to work with the external lib, but not with babel helpers.  They seem to get in-lined regardless of the runtime setting.

jakejrichards: Yeah, I think ideally we would be able to just use `babelHelpers: runtime` everywhere once the bug is fixed and rollup could probably manage to bundle up the imported babel helpers and I don't think this would pollute the global namespace. I think the issue you were running into was probably just the result of [this bug.](https://github.com/rollup/plugins/issues/475).

rollup.config.js
```diff
@@ -24,24 +24,35 @@ const LEGAL_TEXT = `/*
...
+    /@babel\/runtime/ // Need to use a regex here: https://github.com/rollup/plugins/issues/475#issuecomment-652416290
+];
+
+// Need to be able to configure whether or not
+// babel pulls in the esm versions of the runtime helpers
+const babelPlugin = (useESModules = false) => babel({
+    babelHelpers: 'runtime',
+    exclude: /node_modules/,
+    presets: ['@babel/env'],
+    plugins: [['@babel/transform-runtime', { useESModules }]]
```
jakejrichards, line 39: Opted to just leave the babel config in here instead of moving it into a separate `.babelrc` file as it would really only add complexity due to the `useESModules` flag.

PR #72, opened 2020-06-11
ATEAM-272: Migrate to TypeScript

jakejrichards:
This PR gets us from js to ts and replaces `jsdoc2md` with [`typedoc`](https://typedoc.org). I did some local testing using npm link and all of the types seem to be playing nicely. I was also able to plug the docs into the dev-center build script locally and they looked pretty reasonable. Let me know if you want to do this and I can help you set it up.

I suggest reviewing this commit by commit as I tried to keep the scope of each commit limited. 

Some of the guiding principles I followed along the way were:

1. Change as little as possible, only changing the runtime logic if we were doing things that weren't type-safe and they were absolutely necessary for typescript to compile (at least without excessive `as any` casting)

2. Get our outer types to be as strict as possible. There are a few cases where I needed to be a bit lenient. For example, our runtime js currently accepts something like `T | undefined | null`. Ideally, in a second pass, we can limit this to just `T | undefined` to clean up our types a bit.


src/modules/base.ts
```diff
@@ -1,119 +1,138 @@
...
+
+export type MessageListener = (event: SDKMessagePayload) => void;
+export type MessagePayloadFilter = (payload: SDKMessagePayload) => boolean;
+
+interface ListenerOptions {
+    /** 
+     * Set to true to only invoke this listener once; it will be removed after first invocation.
+     * false by default.  null/undefined will use default.
+     */
+    once: boolean;
```
Retsam, line 33: The comment mentions it being optional, but it's not marked as optional (but`msgPayloadFilter` is).   Looks like it's usually used like `Partial<ListenerOptions>`, but maybe it'd be better to do `type ListenerOptions = Partial<{/*...*/}>` instead.

src/modules/base.ts
```diff
@@ -1,119 +1,138 @@
...
+    constructor(cfg: {
+        /** The origin (protocol, hostname, and port) of the target PC environment (e.g. https://apps.mypurecloud.com). Default is '*'. */
+        targetPcOrigin?: string | null,
+        /** The name of the message protocol under which the message will be sent. Default is purecloud-client-apps. */
+        protocol?: string | null;
+        /** The name of the agent from which the message will be sent. Default is purecloud-client-app-sdk (name of the package). */
+        protocolAgentName?: string | null;
+        /** The version of the agent from which the message will be sent. Default is the version of the package. */
+        protocolAgentVersion?: string | null;
+    } = {} as any) {
```
Retsam, line 90: The `as any` cast here looks unnecessary?

src/utils/env.ts
```diff
@@ -11,9 +11,14 @@ const PC_ENV_TLDS = [
...
-let DEFAULT_PC_ENV = null;
-const PC_ENVS = PC_ENV_TLDS.reduce((result, currEnvTld) => {
-    let currEnv = {
+export interface PcEnv {
+    pcEnvTld: string;
+    pcAppOrigin: string;
+}
+
+let DEFAULT_PC_ENV!: PcEnv;
+const PC_ENVS = PC_ENV_TLDS.reduce<PcEnv[]>((result, currEnvTld) => {
```
Retsam, line 20: Sorry, a bit more ⛳ , a `.reduce` that just pushes onto a list looks like a `.map` to me.  Just with a bit of checking to set the `DEFAULT_PC_ENV` variable.

Maybe:

```ts
const PC_ENVS = PC_ENV_TLDS.map((currEnvTld) => ({
     pcEnvTld: currEnvTld,
     pcAppOrigin: `https://apps.${currEnvTld}`
});
// Might consider throwing an error rather than using `!` to assert that it exists.
const DEFAULT_PC_ENV = PC_ENVS.find(({pcEnvTld}) => pcEnvTld === DEFAULT_PC_ENV_TLD)!;
```

tsconfig.json
```diff
@@ -0,0 +1,11 @@
+{
```
Retsam, line 1: In new projects, I like adding `"types"` to the config to specify what `@types` packages to load for global types.  (Otherwise all `@types` packages can contribute global types, even ones that aren't direct project dependencies)  In this case it'd probably be `"types": ["jasmine"]`.  

Looks like that might require some `lib` config; it starts complaining about stuff like `Object.assign` it looks like maybe some dependency is adding those types implicitly?

scripts/build-docs.ts
```diff
@@ -0,0 +1,115 @@
...
+const GITHUB_FORMAT = 'github';
+const PC_DEV_CENTER_FORMAT = 'purecloudDevCenter';
+const SUPPORTED_DOC_OUTPUT_FORMATS = [GITHUB_FORMAT, PC_DEV_CENTER_FORMAT];
+let docMdOutputFormat = process.env.DOC_MD_OUTPUT_FORMAT || PC_DEV_CENTER_FORMAT;
+if (SUPPORTED_DOC_OUTPUT_FORMATS.indexOf(docMdOutputFormat) < 0) {
+    console.error(`Unknown MD Output Format Specified: '${docMdOutputFormat}'`);
+    process.exit(1);
+}
+
+let flagDocsAsPreview = true;
```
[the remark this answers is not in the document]
justinaray: Yeah, a lot of this code was from an era where we had to support IE (maybe 10) 🤮  and I was too lazy to polyfill everything :grin:

Definitely an opportunity to clean up some old junk.  Thanks for pointing it out @Retsam and always feel free.  Love some :golfing:.  And, take them on as needed/desired @jakejrichards.  But, we can also punt any general code clean-up if this PR is getting out of hand.

src/utils/env.ts
jakejrichards: Yeah, this one I ended up changing somewhere off in another PR and my change looks almost exactly like this haha

src/modules/base.ts
jakejrichards: Nice catch 👍

src/modules/base.ts
jakejrichards: Yep, I like this better

scripts/build-docs.ts
```diff
@@ -0,0 +1,115 @@
...
+    excludeProtected: true,
+    excludeNotExported: true,
+    excludeNotDocumented: true
+};
+
+const transformLinks = (buffer: string, ext: string) => {
+    return buffer.replace(/(\[[^\]]+\][^\)]+)(\.md)(\)|\#[^\)]*\))/gm, `$1${ext}$3`);
+};
+
+const prependDocHeaderAttribute = (buffer: string, attr: string) => {
```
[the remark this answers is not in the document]
jakejrichards: Yeah, the generated doc always spits out a header with some predefined fields, but I think its probably a good sanity check since we do have one markdown file that is just copied and not generated.

scripts/build.sh
```diff
@@ -1,4 +1,7 @@
 #!/usr/bin/env bash
 rm -rf dist
+npm run tsc -- --project tsconfig.build.json
+# Rename ts declarations output 'src' -> 'types'
+mv dist/src dist/types
```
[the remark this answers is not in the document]
jakejrichards: Yeah, I was battling this one for a while and finally just settled and decided to do this copy. I'll look into using the environment vars though because that would be much more straightforward 👍

src/ClientAppSpec.ts
```diff
@@ -28,7 +28,7 @@ export default describe('ClientApp', () => {
             let myClientApp = new ClientApp();
             expect(myClientApp.pcEnvironment).toBe(VALID_DEFAULT_PC_ENVIRONMENT);
 
-            myClientApp = new ClientApp({nonEmptyConfig: true});
+            myClientApp = new ClientApp({nonEmptyConfig: true} as any);
```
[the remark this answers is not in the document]
jakejrichards: See, I thought for a bit about how I should handle this one. I was tempted to pull out @ts-expect-error but that seemed a bit wrong as the code itself won't error and there's nothing actually wrong with supplying extra fields on this object as they'll just be ignored. I think `{}` is fair tho as it is its functional equivalent. Unit test casting is always fun 😆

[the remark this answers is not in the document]
Retsam: Yeah, I've started using a bit of the `never` type for stuff like this, where you're deliberately passing incorrect stuff to test how it blows up.  It's basically the same as `any` (can be used in place of any other type), but seems a tiny bit more idiomatic.

jakejrichards:
Thanks for taking a look @Retsam 👍 I really appreciate the feedback!

.eslintrc.js
```diff
@@ -1,21 +1,31 @@
...
             {
                 'anonymous': 'always',
                 'named': 'never'
             }
-        ]
+        ],
+        '@typescript-eslint/ban-types': 'off',
+        '@typescript-eslint/explicit-module-boundary-types': 'off',
+        '@typescript-eslint/no-var-requires': 'off',
+        // TODO: turn these on when project is split into separate src/test folders
```
jakejrichards, line 25: So one thing we should talk about a bit @justinaray is potentially splitting the test files out into their own folder that way we can have separate tsconfig files for each. This would give us the same behavior as before with the environment opt-in for node/jasmine, but with proper types. We could also have separate eslintrc files which could turn some of these more restrictive rules, that we need off for testing, on for source files.

justinaray: @jakejrichards ... yeah, completely agree.  let's move them over to a tests folder when this one lands.

.eslintignore
```diff
@@ -0,0 +1,3 @@
+dist/
```
justinaray, line 1: I've normally seen the targeting of what to lint in the lint script itself vs this ignore, but that doesn't mean this is wrong.  Is this also so you can run eslint from the cli without knowing the incantation of args to run it?

Fine with this, just curious.

jakejrichards: Yeah, the main reason I did it actually is that it's useful for running the vscode eslint plugin so it will ignore these

dev.rollup.config.js
```diff
@@ -37,6 +35,18 @@ const buildExample = (relativeFilePath) => {
     );
 };
 
+const tsc = () => {
+    return new Promise((resolve) => {
+        const bin = require.resolve('typescript/bin/tsc');
+        spawn(bin, [
+            '--incremental',
+            '--outDir', tmpDestPath,
```
justinaray, line 43: Should we output this to a `bin` or `tsbin ` to avoid any overrides?  Not sure if it's really necessary though.

jakejrichards: My initial thought here was that as part of the dev server the compiled declaration files would be output along with the bundles (similar to our actual build process). It's a good idea to output them to a subfolder though. The only reason that we have to output them at all is to allow for incremental builds, which is just an optimization

scripts/build.sh
```diff
@@ -1,4 +1,5 @@
 #!/usr/bin/env bash
```
justinaray, line 1: Should we add `-e` to the shebang so we'll fail fast if tsc:build fails?  The build does ultimately fail if there's a ts compilation issue, but, I think that's only because the docs script is failing and happens to be last.

dev.rollup.config.js
```diff
@@ -1,8 +1,6 @@
-/* eslint-env node */
```
justinaray, line 1: Does the node environment come in from one of the recommended presets by default?  It must as eslint isn't complaining if I reference a node global such as process.  Was just wondering where it comes from.

scripts/build.sh
```diff
@@ -1,4 +1,5 @@
...
 rm -rf dist
```
justinaray, line 2: I can't remember if we talked about this.  Do we want to run the linting/tests as part of the build script either by direct invocation here or via a pre-build script calling `npm run tests tests`?  I see that it's already manually run as part of the proper build, so, I could see us keeping it composable and just opting into it when needed.  I also see it as a pre-push hook, so, it might not be strictly necessary here.

scripts/about.js
```diff
@@ -1,8 +1,4 @@
-#!/usr/bin/env node
```
justinaray, line 1: don't mind, but curious as to why you removed the shebang.  Was something not working?

scripts/build-docs.js
```diff
@@ -1,154 +0,0 @@
...
-    });
-}
-
-/**
- * Returns a copy of the github-flavored markdown srcBuffer transformed into the target format.
- *
- * @param {string} srcBuffer The base github-flavored markdown format
- * @param {string} targetFormat The desired markdwon format of the response
- */
-function transformGithubMarkdown(srcBuffer, targetFormat) {
```
justinaray, line 64: The code blocks look good when running the dev center locally.  Guess that means it now supports gh md code block syntax?

scripts/build-docs.ts
```diff
@@ -0,0 +1,126 @@
...
+    await fs.outputFile(indexFilePath, "---\ntitle: Client App SDK\n---\n\n" + indexFileContents);
+
+    // Apply transformations to docs
+    const docs = await fs.readdir(OUTPUT_DIR);
+    await Promise.all(
+        docs.map(async (doc) => {
+            const docPath = path.join(OUTPUT_DIR, doc);
+            let buffer = (await fs.readFile(docPath)).toString();
+            if (docMdOutputFormat === PC_DEV_CENTER_FORMAT) {
+                buffer = transformLinks(buffer, '.html');
```
justinaray, line 115: Link transform is looking good.  Working between modules and with anchor tags.  The only issue I've seen when running locally is the ./ClientApp.md links in index.html.  Those links seem to be generated with a target="_blank" and open in new tabs.

Not sure if this is an issue with the external link detection algo in dev center or if it's just from running locally.  Maybe @PrinceNebulon could weigh in on this one as well.

scripts/build-docs.ts
```diff
@@ -0,0 +1,126 @@
...
+
+    const project = app.convert(app.expandInputFiles(["src"]));
+    if (!project) throw new Error("Unable to generate typedoc project");
+
+    // Rendered docs
+    app.generateDocs(project, OUTPUT_DIR);
+
+    // Flatten typedoc's generated classes folder
+    const classesDir = path.join(OUTPUT_DIR, "classes");
+    const fileNames = await fs.readdir(classesDir);
```
justinaray, line 90: Will classes always be a flat dir of files?  Just want to make sure a dir wouldn't get copied to dist/docs and then fail on 113 below.

I dont think it's necessary to add a file type guard here at this time if it's always flat though; the build will break and we'll see that.  Just wanted to make sure that was the case now and there weren't any edge cases that would blow us up in the short-term.

src/.eslintrc.js
```diff
@@ -2,8 +2,7 @@ module.exports = {
     parserOptions: {
         sourceType: 'module'
     },
-
     rules: {
-        'no-console': 'error'
+        'no-console': 'error',
```
justinaray, line 6: trailing comma ... nbd tho.

src/ClientAppSpec.ts
```diff
@@ -57,9 +56,9 @@ export default describe('ClientApp', () => {
                 ];
 
                 let index = 0;
-                spyOn(ClientApp, '_getQueryString').and.callFake(() => invalidQueryStrings[++index]);
+                spyOn(ClientApp, '_getQueryString').and.callFake(() => invalidQueryStrings[index++] as any);
```
justinaray, line 59: Fixed a small bug here.  The index was incrementing ahead of the case.  Ironically, when the index eclipsed the size of the array, it would return undefined and still run the intended test.

Was that too much of a happy accident?  Maybe I should have left the code and a comment,
// intentionally exceeding array bounds to test 'undefined' :joy:

src/modules/base.ts
```diff
@@ -1,119 +1,137 @@
...
  *
  * @since 1.0.0
  */
 class BaseApi {
+    private _targetPcOrigin: string;
+    private _protocolDetails: ProtocolDetails;
+
+    // ----- Message Listening
+    private _msgListenerCfgs: Record<string, ListenerConfig[]> = {};
+    private _msgHandler = this._onMsg.bind(this);
```
justinaray, line 56: Will this bind correctly when defined as a property?  I guess we could also use fat arrow functions when registering the listeners as well but I know you are trying to keep the diff down for this PR (and that's appreciated for sure).

Not sure why I didn't just use fat arrows to begin with /shrug

src/modules/base.ts
```diff
@@ -1,119 +1,137 @@
...
-const PROTOCOL_NAME_KEY = 'protocol';
-const PROTOCOL_AGENT_NAME_KEY = 'protocolAgentName';
-const PROTOCOL_AGENT_VERSION_KEY = 'protocolAgentVersion';
+interface ProtocolDetails {
+    name: string;
+    agentName: string;
+    agentVersion: string;
+}
+
+export interface SDKMessagePayload {
```
justinaray, line 16: Thoughts on splitting this interface into and inbound and outboud message payload?

src/modules/base.ts
```diff
@@ -1,119 +1,137 @@
...
+    action: string;
+    hook?: string;
```
justinaray, line 18: If we defined a base interface of InboundSDKMessagePayload here, could we extend it in the classes using listeners so they could add their custom payload details and the base wouldn't need to be aware of them?

Or, would that require making the base class a generic?  If so, we can certainly defer that work as this is the only property/usage as of now and YAGNI.  If we defer, might then make sense to split listeners out as a separate generic class and let downstream classes just compose in the message behavior.

src/modules/base.ts
justinaray: I guess the inbound also technically has a property, `purecloudEventType`, but, if we're not casting to this type and using it, probably not necessary.

scripts/build-docs.ts
PrinceNebulon: Can you provide a specific example? Is there a link on the dev center that's not working correctly currently?

scripts/about.js
jakejrichards: Only reason I removed it was because in the gulp cleanup PR I added an npm script which runs `node scripts/about.js` and so we don't really need to be able to execute this file directly. Open to adding it back tho if desired

scripts/build-docs.js
jakejrichards: Yep, that's the behavior I was seeing so I dropped this.

scripts/build-docs.ts
jakejrichards: Yep, from what I've seen it looks like `classes` is always just a dir of files. The only edge case with our typedoc docs build is if someone went in and added an interface that is both exported and documented then another folder will be generated called `interfaces` alongside `classes`. The docs build will fail in this case as we'll fail when we encounter line 113, but it wouldn't hurt to add something to the README about this and maybe a more specific error.

dev.rollup.config.js
jakejrichards: So the node env is coming from typescript here. In our `tsconfig.json` I explicitly list the included types as `['jasmine']` which excludes node types and that's how I'm able to exclude node globals from our source code. This config doesn't apply to files outside the `src` dir and therefore TS will pull in `@types/node` from node_modules and so eslint will recognize the node globals in the rollup config.

src/modules/base.ts
jakejrichards: Yeah, this is good to consider. It shouldn't require us to make the base generic, so long as the base doesn't need to access these additional properties. I can certainly clean this up a bit now that I understand the structure better.

jakejrichards: Going to correct myself here. It seems like we actually probably do want to go the route of making the base class generic as that will allow us to assert that certain properties are actually on the payload object. My initial thought was that since `hook` was an optional property that it wouldn't require us to make the base class generic. However, after a second go-around in my head, we more than likely will want to assert that it is a required property, and to get TS to be happy with that we will need to make the base generic. So maybe we should punt on this for now. Either way, good to consider.

src/modules/base.ts
jakejrichards: Wasn't seeing any issues with using bind here, but do agree that fat arrow is better.

scripts/build.sh
jakejrichards: Yeah, looking at it now, this `build.sh` file really just turned into a "pre-push" script to ensure all scripts are working. I actually wasn't sure how to make the script fail fast, but `-e` looks like what we need 😄

src/modules/base.ts
```diff
@@ -1,119 +1,137 @@
...
-     * @package
-     * @ignore Extender-use only.  Not a public API
+     * @param actionName
+     * @param msgPayload
      *
      * @since 1.0.0
      */
-    sendMsgToPc(actionName, msgPayload) {
-        this._commsUtils.postMsgToPc(this.buildSdkMsgPayload(...arguments), this._targetPcOrigin);
+    protected sendMsgToPc(actionName: string, msgPayload?: object) {
```
justinaray, line 108: I guess according to existing code, this msgPayload type (and the type in buildSdkMsgPayload) be object | null.  But, since it's protected and the existing usages are either undefined or object, seems reasonable not to widen the type unnecessarily.  Was this your thought as well?

src/modules/base.ts
justinaray: Yeah, this works really nice.  I think the existing js woud have technically allowed null for once but no point in exposing this flexibility to the typed users.  undefined or boolean works perfectly well here.  The normalize will also protect us from js users, so this looks great to me.

src/modules/base.ts
```diff
@@ -1,119 +1,137 @@
...
+    hook?: string;
+    protocol: string;
+    protocolAgentName: string;
+    protocolAgentVersion: string;
+}
+
+export type MessageListener = (event: SDKMessagePayload) => void;
+export type MessagePayloadFilter = (payload: SDKMessagePayload) => boolean;
+
+type ListenerOptions = Partial<{
```
justinaray, line 27: It's not used currently by the lifecycle module (the only module utilizing listeners) but would we want to export this in case an extending class would need to reference the type?

src/modules/base.ts
```diff
@@ -1,119 +1,137 @@
...
 /**
  * Base Class for PureCloud Client App APIs
  *
  * @since 1.0.0
  */
 class BaseApi {
+    private _targetPcOrigin: string;
+    private _protocolDetails: ProtocolDetails;
+
+    // ----- Message Listening
```
justinaray, line 54: Tested lifecycle, once, window message listener cleanup and reactivation when the registered listener count goes to zero and then starts re-adding listeners.  All is looking good 👍

src/modules/baseSpec.ts
```diff
@@ -1,5 +1,4 @@
-/* eslint-env jasmine */
```
justinaray, line 1: Phew, I forgot how extensive these tests were and how much of a PITA it must have been to convert this over to TS.  Thanks!

src/utils/comms.ts
```diff
@@ -3,20 +3,20 @@ export default {
...
+     * @param myConsole - logs errors here rather than throwing errors. Default console
      *
      * @throws Error if the environment is invalid (e.g. not a browser, no postMessage api,
      *  not running within an iframe) and myConsole is not specified
      */
-    postMsgToPc(msg, tgtOrigin, transfer, myWindow = window, myParent = window.parent, myConsole = window.console) {
-        let validRuntime = !!(myWindow && typeof myWindow === 'object' && myParent && typeof myParent === 'object');
-        let validEnv = (validRuntime && myParent !== myWindow);
-        let validApi = !!(myParent && typeof myParent.postMessage === 'function');
+    postMsgToPc(msg: any, tgtOrigin: string, transfer?: Transferable[], myWindow: Window = window, myParent = window.parent, myConsole: Console | null = window.console) {
```
justinaray, line 16: Thanks for pulling in the proper postMessage types!  Looks good.

justinaray: Was going to ask why you make the console nullable and then saw the tests.  I am def fine with it staying be nullable, but we could change the tests to use undefined if you felt strongly about not having it nullable.

Also, not exactly sure how TS deals with it with the fact that impls vary amongst browsers.  For instance, I think <IE10 had window.console undefined if devtools weren't open.

Anyway, maybe best not to open this cat basket?

src/modules/lifecycle.ts
```diff
@@ -1,21 +1,21 @@
...
  */
 
-import BaseApi from './base';
+import BaseApi, { SDKMessagePayload } from './base';
+
+export type LifecycleHook = 'bootstrap' | 'focus' | 'blur' | 'stop';
 
 const LIFECYCLE_HOOK_EVENT_NAME = 'appLifecycleHook';
-const HOOK_NAME_FILTER = function (hookName, msgPayload) {
+const buildHookFilter = (hookName: LifecycleHook) => (msgPayload: SDKMessagePayload) => {
```
justinaray, line 12: Any pro/con to importing MessagePayloadFilter and using that type as the return type for this function?

src/modules/lifecycle.ts
```diff
@@ -74,12 +65,18 @@ class LifecycleApi extends BaseApi {
...
+     * perform post-load initialization work.  This function will be passed the lifecycle event and
+     * does not augment the this context.
+     * @param once - If the listener should only be invoked once or repeatedly; true by default.
      *
      * @see [Advanced Application Concepts - Lifecycle Events](/api/client-apps/advanced.html#lifecycle_events)
      *
      * @since 1.0.0
      */
-    addBootstrapListener(listener, once = true) {
+    addBootstrapListener(listener: (event: SDKMessagePayload) => void, once = true) {
```
justinaray, line 79: Similar question here.  Would it be better to explicitly specify these using the MessageListener type?

src/modules/conversations.ts
```diff
@@ -29,12 +26,13 @@ class ConversationsApi extends BaseApi {
      *         * quality:calibration:view
      *         * quality:evaluation:editScore
      *
-     * @example
```
justinaray, line 32: I noticed you pulled out most of the @example tags but in playing on the tsdoc playground it seems that @example is supported and renders out `Example N`: blocks.

Did you have issues transforming the example blocks to markdown?  Just curious.

justinaray:
Phew @jakejrichards!  What a PR!  Awsome work!  Looking forward to merging this one in the next day or two.  I'm through all the code and we can talk about any/all of these comments.  Mostly nits and questions and overall working perfectly.
👍

src/modules/base.ts
jakejrichards: IMO no need to export until someone actually needs it

src/utils/comms.ts
jakejrichards: Think this could potentially fall under that second pass I described in [this comment](https://github.com/MyPureCloud/client-app-sdk/pull/72#discussion_r473000087) as well. But yeah I'm not sure about IE10 but happy to look into it when we come in to clean this up.

src/modules/lifecycle.ts
jakejrichards: Not opposed here, it really comes down to a slight change in the function signature. Typescript will enforce that these 2 return types to be in sync since one eventually feeds down into the other, but I do like importing since the only reason this guy even exists is to eventually feed down into the listener options.

src/modules/lifecycle.ts
jakejrichards: So yeah, this is probably my least favorite sacrifice that I had to make for typedoc to play nicely. Ideally, we would just import the MessageListener type and use that in all of these public method signatures. When taking this approach, typedoc's generated output only gives you `addBootstrapListener(listener: MessageListener, once: boolean): void` and it won't further expand into what the `MessageListener` type actually is (unless you give it an entire set of pages for interfaces). If you just give the raw type, as is currently done, the docs generate `addBootstrapListener(listener: function, once: boolean): void`, but later on in the parameters output it gives you this type data as well: `(event: SDKMessagePayload): void`.

Let's discuss if we want to open the door to these types which don't actually expose their internal structure in the docs in favor of making the source code simpler.

src/modules/conversations.ts
jakejrichards: Typedoc doesn't support **@example** tags and just renders them like any other tag that it comes across (lowercase bolded text). Their docs suggest just using code blocks as examples. https://typedoc.org/guides/doccomments/#supported-tags. Not that we have to follow that, but I felt that the extra "**example**" text just looked off and detracted from the docs when using the tag in conjunction with the code blocks. We can also discuss this further if you'd like.

src/modules/base.ts
justinaray: Yup, definitely with you on null.  Well, either null or undefined on a case-by-case basis, but def not both.

This repo will be a little annoying as we'll always have to be aware of parse boundaries and check types at the door even though the signatures won't technically allow it.  That said, I'm all for keeping the types tight and also using that as our source of truth for what we mostly expect.

Maybe 3.0 can tighten up the types and we can be less lenient than what we currently allow.

src/modules/lifecycle.ts
justinaray: Ah, yeah, agree that you were between a rock and a hard place.  Having it reference a type that you can't expand upon is no-bueno and opening the door to all interfaces (especially the ones that are private/internal) also seems sub-optimal.  I think we're good with this for now.

Maybe there's a way we can either opt-in/opt-out of interfaces/types in the future.  Maybe we even have to modify a copy of the declaration file or AST to get typedoc to generate what we ultimately want.

But, fine to handle this in pass two.

src/modules/conversations.ts
justinaray: ah gotcha.  Was looking at tsdoc vs typedoc.

Yeah, if it's not formally supported and doesn't provide us a structured block for configuration/styling, I'm def good with just using code blocks.

jakejrichards:
> Great job! LGTM.

Thanks! I appreciate your feedback and all the time it took to get through this one, will be a very satisfying one to ship :shipit:

PR #75, opened 2020-06-23
ATEAM-276: Whitelist lower envs, local dev support, and oauth client id injection

jakejrichards:
The goal here was to move some of our configuration/overrides out of the deploy scripts and into our build. By combining the rollup replace plugin and a few global variable declarations we are able to inject these vars at build time. This makes local development much easier to get started and greatly simplifies our deployment process. I also moved the example transformation/building out into its own script, that way we can reuse it when we deploy the examples.


scripts/build-examples.js
```diff
@@ -0,0 +1,51 @@
...
+const glob = require('glob');
+
+const { CLIENT_APP_SDK_PC_OAUTH_CLIENT_IDS: oauthClientIds } = process.env;
+const BROWSER_FILENAME = `/${pkg.name}.js`;
+
+// Called from command line
+if (!module.parent) {
+    const { bundle } = require('yargs')
+        .option('bundle', { describe: 'minified UMD bundle file name' })
+        .demandOption(['bundle'])
```
justinaray, line 14: Should this be required or should we default it to BROWSER_FILENAME as below in `transformExampleSdkUrl`?  Or, are we assuming this is mostly a feature for builds not local dev and it's not expected users will have to build examples that often?

If we do expect users to build, maybe we should add a blurb about it in README and document the bundle optioon?

src/utils/env.ts
```diff
@@ -8,28 +8,21 @@ const PC_ENV_TLDS = [
...
+}));
 
-    if (currEnvTld === DEFAULT_PC_ENV_TLD) {
-        DEFAULT_PC_ENV = currEnv;
-    }
-
-    result.push(currEnv);
-    return result;
-}, []);
+const [DEFAULT_PC_ENV] = PC_ENVS.filter(env => env.pcEnvTld === DEFAULT_PC_ENV_TLD);
```
justinaray, line 25: Did you use filter over find for browser compat issues?  Works fine, just curious.

jakejrichards: Oh no nothing to do with browser compact, just did this as I feel that it's cleaner since we don't need to do a non-null assertion, as we would with find since it returns `T | undefined`.

scripts/build-examples.js
jakejrichards: Yeah, I think my thought here was that this would be used solely by the Jenkins job, and if someone wants to build/run examples locally then they should probably just start up the dev server, which automatically does the bundling, example transformation, and live reload. With that being said, if we do want to open it up to users, I think a default of BROWSER_FILENAME is reasonable and that we should probably allow them to customize the dist folder as well.

src/index.ts
```diff
@@ -201,6 +188,16 @@ class ClientApp {
         this.externalContacts = new ExternalContactsApi(apiCfg);
     }
 
+    private lookupEnv = (env: string) => {
+        if (env === 'localhost' && __HOST_APP_DEV_ORIGIN__) {
+            this._customPcOrigin = __HOST_APP_DEV_ORIGIN__;
```
justinaray, line 193: This side effect is a bit wonky due to the name `lookupEnv` but it's private and internal so not that big of a deal.  Otherwise, we'd need to complicate the return signature and handle it externally, so, not the end of the world.

src/index.ts
```diff
@@ -201,6 +188,16 @@ class ClientApp {
...
+            return null;
+        }
+        const pcEnv = envUtils.lookupPcEnv(env, true);
+        if (!pcEnv) throw new Error(`Could not parse '${pcEnv}' into a known PureCloud environment`);
```
justinaray, line 197: I think you want `env` in the body of the error message, right?

src/index.ts
```diff
@@ -149,32 +149,19 @@ class ClientApp {
...
-                if (paramValue && typeof paramValue === 'string') {
-                    this._pcEnv = envUtils.lookupPcEnv(paramValue, true);
-                    if (!this._pcEnv) {
-                        throw new Error(`Could not parse '${paramValue}' into a known PureCloud environment`);
-                    }
-                } else {
-                    throw new Error(`Could not find unique value for ${paramName} parameter on Query String`);
-                }
-            } else if (cfg.hasOwnProperty('pcEnvironment')) {
+                this._pcEnv = this.lookupEnv(parsedQueryString[paramName]);
```
justinaray, line 158: The queryString lib can actually return a string or a string[] depending if their are duplicate keys.

Looks like envUtils.lookupPcEnv will fail out if the type passed is not string, but seems a bit brittle to rely on this.  Should we attempt to narrow the type here and fail if it's not a string, or, were you ok with the pass through and fail from envUtils?

scripts/build-examples.js
justinaray: Yeah, i'm good with keeping it as primarily a deploy function.  Maybe a note in the readme describing the local server and examples vs the deployed examples would be good ... but only if we have to make another commit to this PR for something less nitty.

scripts/build-docs.ts
```diff
@@ -1,27 +1,22 @@
...
-            flagDocsAsPreview = false;
-        }
-    });
-}
+const SUPPORTED_DOC_OUTPUT_FORMATS = ['github', 'purecloudDevCenter'] as const;
+
+const { preview: flagDocsAsPreview, format: docMdOutputFormat } = yargs
+    .option('preview', {
+        type: 'boolean',
+        default: false
```
justinaray, line 13: Conscious decision to switch the default?  Seems reasonable give we're live but wanted to make sure.

scripts/build-docs.ts
```diff
@@ -1,27 +1,22 @@
 import * as path from "path";
 import * as fs from "fs-extra";
 import { Application, TSConfigReader, TypeDocOptions } from "typedoc";
+import * as yargs from 'yargs';
```
justinaray, line 4: Nice!  So much easier to deal with!

karma.conf.js
```diff
@@ -39,16 +52,18 @@ module.exports = function (config) {
         concurrency: Infinity,
         rollupPreprocessor: {
             plugins: [
-                require('@rollup/plugin-replace')({
-                    '__PACKAGE_NAME__': JSON.stringify(process.env.npm_package_name),
-                    '__PACKAGE_VERSION__': JSON.stringify(process.env.npm_package_version)
+                replace({
+                    '__PACKAGE_NAME__': JSON.stringify(npm_package_name),
+                    '__PACKAGE_VERSION__': JSON.stringify(npm_package_version),
+                    '__HOST_APP_DEV_ORIGIN__': JSON.stringify(devOrigin),
```
justinaray, line 58: Any issues with this getting replaced as `undefined`?  Seems like it will get replaced inline and, at most, it would be a warning that the expression is constant.

`if (env === 'localhost' && __HOST_APP_DEV_ORIGIN__) {`

package.json
```diff
@@ -9,7 +9,7 @@
   "files": [
     "/LICENSE",
     "/README.md",
-    "/dist/*.*",
+    "/dist/",
```
justinaray, line 12: this won't include the dist dir itself, right?

scripts/tsconfig.json
```diff
@@ -0,0 +1,10 @@
+{
+    "compilerOptions": {
+        "module": "commonjs",
```
justinaray, line 3: default target looks like es3 if not specified.  But, shouldn't matter for examples.

scripts/build-docs.ts
jakejrichards: Yeah, pretty sure it was intentional / think it makes reasonable sense that we would have to explicitly mark as a "preview" version.

src/index.ts
jakejrichards: Lol, this PR is approaching the length of time ago where I'm questioning what past me was thinking 😆 Talk about wonk...

package.json
justinaray: nvmd ... it was there and remains there.  nothing to see here 🤦‍♂️

karma.conf.js
jakejrichards: Shouldn't be any issues, it will evaluate to `""`, which is falsy and we'll just bypass the block and carry on without checking for a custom localhost origin.

scripts/tsconfig.json
jakejrichards: Yeah, it's actually for scripts and the scripts are only ever run with `ts-node` which compiles and executes for us so we shouldn't need to dictate a newer target.

src/index.ts
jakejrichards: Alright, this wonkiness and the string type check for the parsed query param value are fixed in [this commit](https://github.com/MyPureCloud/client-app-sdk/pull/75/commits/697bc803d2e76b8e1823159cd6a9f6ffbd85020d).

CONTRIBUTING.md
```diff
@@ -32,22 +32,62 @@ $ yarn run watch:test
...
+$ yarn run examples --bundle={BUNDLE_PATH}
 ```
 
 ## Contributing
 1. Fork the repo
 1. Add your code (Don't forget the unit tests!)
 1. Test your code
   * `npm test` -or- `npm run watch:test`
-  * `npm run build` and try the version in an app -or- `npm run serve` and try the example in PC
+  * `npm run build` and try the version in an app -or- `npm start` and try the example in PC
```
justinaray, line 69: nit: GC

CONTRIBUTING.md
```diff
@@ -32,22 +32,62 @@ $ yarn run watch:test
...
 
 # Starts a test server with the examples using the local sdk build. Watches for changes.
-$ yarn run serve 
+$ yarn start 
 # -or-
-$ npm run serve
+$ npm start
+```
+
+## Examples
```
justinaray, line 40: :+1: This section is perfect and a big help.

src/index.ts
```diff
@@ -149,36 +149,18 @@ class ClientApp {
...
-                if (!this._pcEnv) {
-                    throw new Error(`Could not parse '${cfg.pcEnvironment}' into a known PureCloud environment`);
-                }
-            } else if (cfg.hasOwnProperty('pcOrigin')) {
-                if (typeof cfg.pcOrigin !== 'string' || cfg.pcOrigin.trim().length === 0) {
-                    throw new Error('Invalid pcOrigin provided.  Must be a non-null, non-empty string');
-                }
-
+                this.assertNonEmptyString(paramValue, 'query param value');
+                this._pcEnv = this.lookupEnv(paramValue);
```
justinaray, line 158: I think this creates a change in behavior, right?

As lookupEnv doesn't throw, it will be null and then overridden by the default?  I think what we had (and still want?) is to fail out if the value provided isn't valid, right?

src/index.ts
```diff
@@ -149,36 +149,18 @@ class ClientApp {
...
-                this._pcEnv = envUtils.lookupPcEnv(cfg.pcEnvironment, true);
-                if (!this._pcEnv) {
-                    throw new Error(`Could not parse '${cfg.pcEnvironment}' into a known PureCloud environment`);
-                }
-            } else if (cfg.hasOwnProperty('pcOrigin')) {
-                if (typeof cfg.pcOrigin !== 'string' || cfg.pcOrigin.trim().length === 0) {
-                    throw new Error('Invalid pcOrigin provided.  Must be a non-null, non-empty string');
-                }
-
+                this.assertNonEmptyString(paramValue, 'query param value');
```
justinaray, line 157: you could use the param name here for a more obvious error message.  Wouldn't have worried about this nit here, but you might have to get back into this to fix the below comment anyway.

src/index.ts
justinaray: ugh ... I'm an idiot.  nvmd

justinaray (approved):
Looks really good @jakejrichards ... so much easier now!  Nice work!

PR #76, opened 2020-07-09
ATEAM-281: Preserve pcEnvironment query param after oauth flow

jakejrichards:
So after some digging, it appears that this issue has always been around and that if you reload the frame on any examples that use the implicit login flow you'll run into issues where the `pcEnvironment` query param will no longer be found. This can be worked around by just hardcoding your `pcEnvironment` in the html/js file, but I like not having to do any setup when starting the dev server being able to start up without any magic or special hidden config/workarounds. Here's why this solves the problem:

1. The URL used in the configuration of an app is `https://localhost:8443/profile.html?pcEnvironment={{pcEnvironment}}`
2. App initially loads at `https://localhost:8443/profile.html?pcEnvironment=mypurecloud.com`
3. App initiates implicit login flow and is redirected after authentication to `https://localhost:8443/profile.html#access_token=...&state=pcEnvironment%253Dmypurecloud.com&token_type=...`
4. We parse the `pcEnvironment` from the hash and the platform client then [parses the access token and clears the hash](https://github.com/MyPureCloud/platform-client-sdk-javascript/blob/708792a0557ddabb6df7a7455b2422d2c5f2bf5f/build/src/purecloud-platform-client-v2/ApiClient.js#L460)
5. Then, after all is done and we are authenticated, I am now manually restoring the query param using `window.history.replaceState` which is an API that allows you to modify the history without reloading the window.

Also, while in here I pulled out a common `authenticate` utility function which is now used in all of the examples where we need to login with the implicit oauth flow.

### DIFF
--- a/dev.rollup.config.js
+++ b/dev.rollup.config.js
@@ -24,7 +24,7 @@ try {
 const transformExampleSdkUrl = (buffer) => {
     return buffer.replace(
         /(\s*<script.*src=")([^"]+client-app-sdk[^"]+)(".*<\/script>\s*)/i,
-        '$1' + (BROWSER_FILENAME || '$2') + '$3\n'
+        '$1' + (`/${BROWSER_FILENAME}` || '$2') + '$3\n'
     );
 };
 
--- a/examples/agentInteractionsDemo/app.js
+++ b/examples/agentInteractionsDemo/app.js
@@ -158,19 +158,6 @@ new Vue({
             return;
         }
 
-        /*
-            * Note: To use this app in your own org, you will need to create your own OAuth2 Client(s)
-            * in your PureCloud org.  After creating the Implicit grant client, map the client id(s) to
-            * the specified region key(s) in the object below, deploy the page, and configure an app to point to that URL.
-            */
-        let pcOAuthClientIds = {'mypurecloud.com': 'implicit-oauth-client-id-here'};
-
-        let clientId = pcOAuthClientIds[pcEnvironment];
-        if (!clientId) {
-            this.errorMessage = pcEnvironment + ': Unknown/Unsupported PureCloud Environment';
-            return;
-        }
-
         let client = platformClient.ApiClient.instance;
         client.setEnvironment(pcEnvironment);
         client.setPersistSettings(true);
@@ -198,17 +185,9 @@ new Vue({
         let authenticated = false;
         let agentUserId = "";
 
-        let redirectUrl = window.location.origin;
-        if (!redirectUrl) {
-            redirectUrl = window.location.protocol + '//' + window.location.host;
-        }
-        redirectUrl += window.location.pathname;
-
         // Authentication and main flow
-        client.loginImplicitGrant(clientId, redirectUrl, { state: ("pcEnvironment=" + pcEnvironment) })
-            .then((data) => {
-                window.history.replaceState(null, null, window.location.pathname + "?" + data.state);
-                // Get userme info
+        authenticate(client, pcEnvironment)
+            .then(() => {
                 authenticated = true;
                 return usersApi.getUsersMe({ "expand": ["presence"] });
             })
@@ -268,8 +247,10 @@ new Vue({
             })
             .catch((err) => {
                 console.log(err);
-                this.errorMessage = !authenticated ? "Failed to Authenticate with PureCloud" :
-                                    "Failed to fetch/display profile";
+                this.errorMessage =
+                    !authenticated
+                        ? "Failed to Authenticate with PureCloud - " + err.message
+                        : "Failed to fetch/display profile";
             });
     },
 });
--- a/examples/agentInteractionsDemo/app.js
+++ b/examples/agentInteractionsDemo/app.js
@@ -206,7 +206,8 @@ new Vue({
 
         // Authentication and main flow
         client.loginImplicitGrant(clientId, redirectUrl, { state: ("pcEnvironment=" + pcEnvironment) })
-            .then(() => {
+            .then((data) => {
+                window.history.replaceState(null, null, window.location.pathname + "?" + data.state);
                 // Get userme info
                 authenticated = true;
                 return usersApi.getUsersMe({ "expand": ["presence"] });
--- a/examples/interactionsDemo.html
+++ b/examples/interactionsDemo.html
@@ -21,6 +21,8 @@
 
         <script src="https://sdk-cdn.mypurecloud.com/javascript/72.0.0/purecloud-platform-client-v2.min.js"></script>
         <script src="https://sdk-cdn.mypurecloud.com/client-apps/1.4.0/purecloud-client-app-sdk-49c1570bfd8967d8861435e0e7178035.min.js"></script>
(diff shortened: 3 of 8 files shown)


dev.rollup.config.js
```diff
@@ -26,7 +26,7 @@ try {
 const transformExampleSdkUrl = (buffer) => {
     return buffer.replace(
         /(\s*<script.*src=")([^"]+client-app-sdk[^"]+)(".*<\/script>\s*)/i,
-        '$1' + (BROWSER_FILENAME || '$2') + '$3\n'
+        '$1' + (`/${BROWSER_FILENAME}` || '$2') + '$3\n'
```
jakejrichards, line 29: Noticed this tiny bug while in here, need to prepend a `/` to the path otherwise the examples with a nested folder structure won't be able to find the sdk.

examples/utils/oauth.js
```diff
@@ -0,0 +1,21 @@
+// Authenticate with PureCloud
+function authenticate(client, pcEnvironment) {
+    /*
+    * Note: To use this app in your own org, you will need to create your own OAuth2 Client(s)
+    * in your PureCloud org.  After creating the Implicit grant client, map the client id(s) to
+    * the specified region key(s) in the object below, deploy the page, and configure an app to point to that URL.
+    */
+    const pcOAuthClientIds = {'mypurecloud.com': 'implicit-oauth-client-id-here'};
```
jakejrichards, line 8: Nice that now this only lives in one place, especially when we do the injection later in the new build process.

examples/utils/oauth.js
```diff
@@ -0,0 +1,21 @@
+// Authenticate with PureCloud
+function authenticate(client, pcEnvironment) {
```
jakejrichards, line 2: We could alternatively just pull the platform client instance from the window (i.e. `window.require('platformClient')`, but different examples with setting different configurations on the client instance and it felt a little magical to just pull the same instance down from the window and this is more explicit. I'm open to going this route tho as we would probably be able to eliminate a tiny bit of duplication.

justinaray: I think you made the right call.  I'll always favor explict over magical.  Also, I think this is the right balance of showing some config but hidding other nitty gritty details like auth in a util to make the demo code easier to read.

examples/agentInteractionsDemo/app.js
```diff
@@ -158,19 +158,6 @@ new Vue({
...
-        /*
-            * Note: To use this app in your own org, you will need to create your own OAuth2 Client(s)
-            * in your PureCloud org.  After creating the Implicit grant client, map the client id(s) to
-            * the specified region key(s) in the object below, deploy the page, and configure an app to point to that URL.
-            */
-        let pcOAuthClientIds = {'mypurecloud.com': 'implicit-oauth-client-id-here'};
-
-        let clientId = pcOAuthClientIds[pcEnvironment];
-        if (!clientId) {
-            this.errorMessage = pcEnvironment + ': Unknown/Unsupported PureCloud Environment';
```
justinaray, line 170: Looks like all of these examples were already including the dynamic pcEnvironment value in their error messages.  The frameworks take care of it, but I'm always cognizant of XSS when we're using raw dom apis.  I believe I caught all the innerHTMLs, etc. when the examples went through initial PR, but we should give this a once over just to confirm.

examples/interactionsDemo.html
```diff
@@ -214,12 +198,12 @@ <h2>Manual Tester</h2>
                 let canViewInteraction = false;
                 let canViewWrapUpCodes = false;
 
-                client.loginImplicitGrant(clientId, redirectUrl, { state: ('pcEnvironment=' + pcEnvironment) })
-                    .then(function () {
+                authenticate(client, pcEnvironment)
+                    .then(() => {
                         authenticated = true;
-                        return new platformClient.UsersApi().getUsersMe({ 'expand': ['authorization'] });
+                        new platformClient.UsersApi().getUsersMe({ 'expand': ['authorization'] })
```
justinaray, line 204: Was this an intentional removal of the return statement?  Don't we want to wait on the resolution of this promise for the next then block?

jakejrichards: Nice catch 👍 Definitely need to return here. Just looked at this a bit and funny enough, since we only do the boolean flag checking, we fail silently here and so I wasn't seeing errors. `profileData` is undefined on 206 and we try to read `profileData.authorization.permissions` on 210 which throws a type error, but since we set both `authenticated` and `userDataAcquired` to true. The catch block just ignores the error and hides the loader as if nothing happened. Anyway, will get this fixed this up 😄

examples/utils/oauth.js
```diff
@@ -0,0 +1,21 @@
...
+    if (!clientId) {
+        return Promise.reject(new Error(pcEnvironment + ': Unknown/Unsupported PureCloud Environment'));
+    }
+
+    const { origin, protocol, host, pathname } = window.location;
+    const redirectUrl = (origin || `${protocol}//${host}`) + pathname;
+
+    return client.loginImplicitGrant(clientId, redirectUrl, {state: `pcEnvironment=${pcEnvironment}`})
+        .then(data => {
+            window.history.replaceState(null, null, `${pathname}?${data.state}`);
```
justinaray, line 19: mdn suggests using '' for the title as opposed to null for future proofing the call across browsers.  But, I haven't used history state much, so take this with a grain of salt if null is more appropriate.

justinaray: Also, nice approach using replaceState.  Clean, simple, addresses the issue 👍

examples/utils/oauth.js
```diff
@@ -0,0 +1,21 @@
...
+    const pcOAuthClientIds = {'mypurecloud.com': 'implicit-oauth-client-id-here'};
+    const clientId = pcOAuthClientIds[pcEnvironment];
+    if (!clientId) {
+        return Promise.reject(new Error(pcEnvironment + ': Unknown/Unsupported PureCloud Environment'));
+    }
+
+    const { origin, protocol, host, pathname } = window.location;
+    const redirectUrl = (origin || `${protocol}//${host}`) + pathname;
+
+    return client.loginImplicitGrant(clientId, redirectUrl, {state: `pcEnvironment=${pcEnvironment}`})
```
justinaray, line 17: Might need to expand this to an object in the future so callers can serialize more data, but we'll cross that bridge when we get to it.  Looks good for now.

dev.rollup.config.js
justinaray: Ah, nice catch!

examples/agentInteractionsDemo/app.js
justinaray: Confirmed that all of these examples just present it as text, not html.  So, g2g

examples/interactionsDemo.html
justinaray: Doh, guess we need an else with a generic message, huh?  Thanks for looking into it!

justinaray (approved):
Nice work as usual!  replaceState is a clean fix to this and also cleaned up a lot of duplicated code.

PR #78, opened 2020-07-28
ATEAM-111: Add Propose Message Handler + Demo

jakejrichards:
The purpose of this SDK method is to fill the agent's active interaction text box. This method will only work with the currently active interaction and when the calling app is the active app. This is to avoid the agent's textbox being changed from multiple sources and so that the agent is aware when the text content has changed.

I mostly just worked off of what Eric had already done here and cleaned up a few things. Also, I branched off of the build-transformations branch here so that I was able to deploy this demo.

### DIFF
--- a/examples/proposeMessageDemo.html
+++ b/examples/proposeMessageDemo.html
@@ -1,0 +1,102 @@
+<!DOCTYPE html>
+<html>
+    <head>
+        <title>Genesys Cloud Propose Message Demo</title>
+        <script src="https://sdk-cdn.mypurecloud.com/client-apps/1.4.0/purecloud-client-app-sdk-49c1570bfd8967d8861435e0e7178035.min.js"></script>
+        <style>
+            #app {
+                display: flex;
+                flex-direction: column;
+                max-width: 300px;
+            }
+            .button-list, form {
+                display: flex;
+                flex-direction: column;
+                border: 1px solid #ccc;
+                padding: 0 5px;
+            }
+            button, input, div {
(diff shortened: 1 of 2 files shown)


src/modules/conversations.ts
```diff
@@ -35,6 +35,20 @@ class ConversationsApi extends BaseApi {
     showInteractionDetails(conversationId: string) {
         super.sendMsgToPc('showInteractionDetails', {conversationId: conversationId});
     }
+
+    /**
+     * Send a message to be filled into the interaction message box for the agent to review and send.
+     * This function works specifically with a bound interaction when both the interaction and calling app
+     * are visible, it is not intended (and will not work) for situations where the interaction is not active.
```
justinaray, line 42: Can you add a sentence to the effect of,

"Practically, this means that you should only call this method in response to user interaction to ensure the agent is aware of the impending text insertion and so their existing draft state is not unexpectedly altered."

justinaray (approved):
looks great @jakejrichards ... thanks!

examples/proposeMessageDemo.html
```diff
@@ -0,0 +1,102 @@
...
+                cursor: not-allowed;
+            }
+        </style>
+    </head>
+    <body>
+        <script>
+        (function () {
+        const clientApp = new window.purecloud.apps.ClientApp({ pcEnvironmentQueryParam: 'pcEnvironment' });
+
+        const html = (htmlStr, template = document.createElement('template')) => {
```
justinaray, line 51: :fire:

PR #79, opened 2020-07-29
ATEAM-287: Conditionally generate a set of html or genesys cloud dev-center docs

jakejrichards:
As @justinaray and I discussed the other day, in addition to building docs specifically for the dev center, we'd like to be able to generate pure set of docs that can be hosted elsewhere. This PR splits our doc generation process based on what format the user specifies. The way it's set up now we should be able to handle HTML and dev-center output formats. We can always choose to add a pure markdown output or something of the sort if we choose later on. This just gives us the initial setup. While I was here I also added a clean step to run before docs generation.


jakejrichards:
Wrong ticket number, will re-open

PR #80, opened 2020-07-29
ATEAM-287: Conditionally generate a set of html or genesys cloud dev-center docs

jakejrichards:
As @justinaray and I discussed the other day, in addition to building docs specifically for the dev center, we'd like to be able to generate pure set of docs that can be hosted elsewhere. This PR splits our doc generation process based on what format the user specifies. The way it's set up now we should be able to handle HTML and dev-center output formats. We can always choose to add a pure markdown output or something of the sort if we choose later on. This just gives us the initial setup. While I was here I also added a clean step to run before docs generation.


scripts/docs/build-html-docs.ts
```diff
@@ -0,0 +1,28 @@
...
+export const buildHtmlDocs = async ({ outputDir, options }: HtmlDocsOptions) => {
+    const app = new Application();
+
+    // So TypeDoc can load tsconfig.json
+    app.options.addReader(new TSConfigReader());
+
+    app.bootstrap({
+        ...options,
+        theme: 'minimal',
+        readme: 'none',
```
justinaray, line 17: nit:  did you consider adding docs/index.md as the index file?  I tried it and it seems to work, but, we'd need to mutate the relative ./ClientApp.md links here as well as we do in the dev-center.

Just curious on your thoughts here.  It's a nit and we def don't need to add it as we don't have a consumer of the html docs at this point.

scripts/docs/build-dev-center-docs.ts
```diff
@@ -0,0 +1,112 @@
...
+    return buffer = `---\n${attr}: ${value}${buffer.substring('---'.length)}`;
+};
+
+const injectPackageVersion = (version: string) => (buffer: string) => {
+    const result = buffer.replace(/<taggedversion>/g, version);
+    return prependDocHeaderAttribute('version', version)(result);
+};
+
+const injectBundleHash = (bundleHash: string) => (buffer: string) => {
+    const result = buffer.replace(/<hash>/g, bundleHash);
```
justinaray, line 110: This has been long requested!  Thanks!

scripts/docs/build-dev-center-docs.ts
```diff
@@ -0,0 +1,112 @@
...
+    return prependDocHeaderAttribute('bundle_hash', bundleHash)(result);
```
justinaray, line 111: Is this consumed in any way?  Or just metadata?

scripts/docs/build-dev-center-docs.ts
```diff
@@ -0,0 +1,112 @@
...
+    );
+
+    // Delete typedoc's generated classes folder
+    await fs.remove(classesDir);
+
+    // Use custom index file in place of typedoc's index
+    await fs.copy("doc/index.md", path.join(outputDir, "index.md"));
+
+    // Add header to index file
+    await transformFiles(outputDir, buffer => "---\ntitle: Client App SDK\n---\n\n" + buffer);
```
justinaray, line 68: Looks like we're actually adding the title to every doc file.  Any issue with this with devCenter?  If so, we'd need something else here so prepend will work.  It assumes a leading `---`.

justinaray: Looks like the generator creates some metadata automatically.  So, if we add this initial block to all files unconditionally, looks like we create multiple metadata blocks.

jakejrichards: Whooops, not sure how that happened. This was supposed to only have been run on the index file. But I def now see that it's running on everything 😅

scripts/docs/build-dev-center-docs.ts
jakejrichards: Just metadata

scripts/docs/build-dev-center-docs.ts
jakejrichards: Should be g2g here now 👍  @justinaray

justinaray (approved):
Looks good! Really like the pluggability of the transforms!

PR #81, opened 2020-07-29
ATEAM 283: Improve toast error messages and docs

carolinepar:


jakejrichards:
> Should I also include "toast messages of types other than 'error' require either an automatic dismiss or a close button"? If so, should I change line 23 to "Displays a toast popup. Toasts with type ‘error’ do not allow automatic dismissal and require a close button. Toast messages of types other than 'error' require either an automatic dismiss or a close button." and remove the change to line 33?

I think I'm in favor of updating the docs on line 23 as you said. That way we give a conceptual overview of what consumers are expected to supply. Although one small nit for the second sentence, perhaps we could say: "Toast messages that require manual dismissal (i.e. a timeout of `0`) are required to show the close button."

I think this actually something that we can enforce via typescript (when that's merged) using type params. I think something like this would get the job done:
```typescript
type ToastOptions<Type extends 'error' | 'info' | 'success', Timeout extends number> = {
    type: Type;
    timeout: Type extends 'error' ? 0 : Timeout;
    showCloseButton: Timeout extends 0 ? true : boolean;
}
```

src/modules/alerting.ts
```diff
@@ -42,20 +52,30 @@ class AlertingApi extends BaseApi {
...
+     *
+     * Error toasts (`type: 'error'`) require manual dismissal and must be specified with `showCloseButton: true`.
+     * Any specified `timeout` will be ignored.
+     *
      * ```ts
      * myClientApp.alerting.showToastPopup("Hello world", "Hello world, how are you doing today?");
      * ```
      *
      * ```ts
      * var options = {
```
jakejrichards, line 67: Updated our examples while I was here to reflect our newly enforced contract (at least for TS consumers)

src/modules/alerting.ts
```diff
@@ -32,6 +26,22 @@ const pick = <
...
+type ToastOptions<Type extends ValidMessageType, Timeout extends number> = {
+    /**
+     * The id of the message.  Duplicate IDs will replace each other in the toast display.  All IDs will be
+     * namespaced with your app ID to avoid collisions.  (Default will just be your app's namespace and will not support multiple messages)
+     * */
+    id?: string;
+    /** Toast type, valid options are 'error', 'info', 'success'.  (Default is 'info') */
+    type: Type;
+    /** Time in seconds to show the toast.  Set to 0 to disable automatic dismissal. (Default is 5) */
+    timeout: Type extends 'error' ? 0 : Timeout;
```
justinaray, line 38: Makes perfect sense now, but when we were chatting about this, I didn't think about the implication that type, timeout, and showCloseButton are now required.  This makes the using the options a bit clumsy as you can't just bump the timeout with `{timeout: 10}`.

src/modules/alerting.ts
```diff
@@ -68,33 +88,29 @@ class AlertingApi extends BaseApi {
...
-        showCloseButton?: boolean;
-    } = {}) {
-        const messageParams: BaseToastMessageProps & typeof options = {
+    showToastPopup<
+        MessageType extends 'error' | 'info' | 'success',
+        Timeout extends number
+    >(
+        title: string,
+        message: string,
+        options: ToastOptions<MessageType, Timeout> = {} as ToastOptions<MessageType, Timeout>
```
justinaray, line 111: Should we go ahead and specify a default as `{type: 'info', timeout: 5, showCloseButton: false}`?

src/modules/alerting.ts
```diff
@@ -42,13 +72,20 @@ class AlertingApi extends BaseApi {
...
+     * Any specified `timeout` will be ignored.
+     *
      * ```ts
      * myClientApp.alerting.showToastPopup("Hello world", "Hello world, how are you doing today?");
      * ```
      *
      * ```ts
      * var options = {
-     *    type: 'success'
+     *    type: 'success',
```
justinaray, line 88: nit: dangling comma

src/modules/alerting.ts
```diff
@@ -72,29 +109,22 @@ class AlertingApi extends BaseApi {
...
-        /** Boolean indicating if the message is in MD.  (Default is true) */
-        markdownMessage?: boolean;
-        /** Time in seconds to show the toast.  Set to 0 to disable automatic dismissal. (Default is 5) */
-        timeout?: number;
-        /** Boolean indicating if the close button should be shown. (Default is false) */
-        showCloseButton?: boolean;
-    } = {}) {
-        const messageParams: BaseToastMessageProps & typeof options = {
+    showToastPopup<
+        MessageType extends 'error' | 'info' | 'success',
```
justinaray, line 120: Should this extend ValidMessageType here?

src/modules/alerting.ts
justinaray: Digging into this more, the defaults are actually showCloseButton: true, timeout: 7.  So, changing this would actually be a "breaking change for 2.0".  Wondering if we should just update default docs to match behavior and move on.

src/modules/alerting.ts
```diff
@@ -32,6 +26,42 @@ const pick = <
     return newObj;
 };
 
+type ToastOptions<Type extends ValidMessageType, Timeout extends number> = {
```
justinaray, line 29: This is working great @jakejrichards!  Optional props are back and like the enhanced descriptions in the type errors!

src/modules/alerting.ts
jakejrichards: So what I've got here `{}` for the default matches what we had before correct?

src/modules/alerting.ts
jakejrichards: Ahh yeah, good catch

src/modules/alerting.ts
justinaray: Yeah, it's the same as before.  The problem is that the documented defaults are (and may have always been) incorrect.

Maybe we should handle this in another PR?  that being we should enforce/set our specified defaults here so they're not affected by downstream changes?

src/modules/alerting.ts
```diff
@@ -122,7 +122,12 @@ class AlertingApi extends BaseApi {
     >(
         title: string,
         message: string,
-        options: ToastOptions<MessageType, Timeout> = {} as ToastOptions<MessageType, Timeout>
+        options: ToastOptions<MessageType, Timeout> = {
+            type: 'info',
+            timeout: 7,
+            showCloseButton: false,
+            markdownMessage: true
+        } as unknown as ToastOptions<MessageType, Timeout>
```
jakejrichards, line 130: So this was an unfortunate cast I had to make to get TS to play nicely. But the most important thing is that we're typesafe from the outside - and we still are here. The heart of the issue is that the consumer of this method can specify a subtype of `ValidMessageType` (MessageType type param) or `number` (Timeout type param) that doesn't intersect with `'info'` or `7` so TS can't assert that this is a safe default.

justinaray: For a minute, I was considering if we should move all these props down to messageParams and overlay the values.  That would let us provid prop defaults even when an options partial was provided by the sdk user.  However, that would be breaking behavior as showCloseButton is evaluated for falsey and undefined would be different than our newly specified default (even if our new defaults are now functionally correct)

I'm assuming that most users are now used to the quirks of this options hash at this point and they're already specifying the props they need.  So, tl;dr, I think we can just keep this as a fully default param and not break the internals too much.

In the future, I think the best path forward might be to provide a more explicit api under different method(s) and deprecate this one; like we discussed.

justinaray: image

justinaray: Made one last pass at the docs to try to match our behavior.  Take a look and keep me honest.

justinaray (approved):

src/modules/alerting.ts
jakejrichards: LGTM :shipit:

PR #82, opened 2020-08-13
DEVCTR-121: Genesys Cloud Rebranding

jakejrichards:


package.json
```diff
@@ -34,9 +34,9 @@
     "build"
   ],
   "keywords": [
```
jakejrichards, line 36: Not sure if we should keep the old keywords as well for search purposes, figured they'd be okay to drop, but can add them back in if needed.

examples/urlInterpolationDemo.html
```diff
@@ -42,21 +42,21 @@ <h2>Localization Demo</h2>
                 var DEFAULT_LANG_TAG = 'en-us';
                 var langTag = null;
                 var welcomeTokens = {
```
jakejrichards, line 44: Not sure if these were localized by the localization team, but figured this change was still reasonable.

justinaray: Yeah, I think I picked these up from a bundle somewhere just to illustrate.  Should be good to change here.

package.json
justinaray: Hmm ... maybe keep the MyPureCloud since that's still the organization name for the time being.

justinaray: so, mypurecloud, genesys cloud, genesys cloud apps

PR #86, opened 2020-11-30
DEVCTR-121: Genesys Cloud Rebranding

jakejrichards:
This is the PR to rebrand our docs & examples from PureCloud -> Genesys Cloud. This should be included with our 2.0 release that way we can publish these changes as part of the new typedoc generated docs.

### DIFF
--- a/src/utils/comms.ts
+++ b/src/utils/comms.ts
@@ -1,7 +1,7 @@
 export default {
     /**
      * Posts the provided msg to the specified origin and parent window when invoked within the
-     * expected environment (e.g. a frame hosted within PureCloud).
+     * expected environment (e.g. a frame hosted within Genesys Cloud).
      *
      * @param msg - The payload to send as a postMessage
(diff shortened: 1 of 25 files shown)


carolinepar (approved):

src/utils/comms.ts
```diff
@@ -21,7 +21,7 @@ export default {
         if (validRuntime && validEnv && validApi) {
             myParent.postMessage(msg, tgtOrigin, transfer);
         } else {
-            let errMsg = 'PureCloud Communication Failed.  ';
+            let errMsg = 'Genesys Cloud Communication Failed.  ';
```
justinaray, line 24: Let's switch this one back just on the off chance a client is using the string in an error check.  This could be thrown by any of the sdk methods.

And we'll get this one with the major release where where we switch the artifact from purecloud to genesysCloud.

justinaray (approved):
Looks good @jakejrichards ... Good to ship once the one reference is reverted in src/utils/comms.ts

PR #87, opened 2020-12-01
Feature/ateam 309 audit app docs

carolinepar:


jakejrichards (approved):

README.md
```diff
@@ -85,7 +85,7 @@ PUT /api/v2/integrations/{integrationInstanceId}/config/current
     "version": {currAppConfigVersion},
     "properties" : {
         "url" : "https://mypurecloud.github.io/client-app-sdk/help.html",
-        "sandbox" : "allow-forms,allow-modals,allow-popups,allow-presentation,allow-same-origin,allow-scripts",
+        "sandbox" : "allow-forms,allow-modals,allow-popups,allow-presentation,allow-same-origin,allow-scripts,allow-downloads",
```
justinaray, line 88: Nice catch!

examples/urlInterpolationDemo.html
```diff
@@ -83,8 +84,12 @@ <h2>Localization Demo</h2>
                         pcEnv = getParamValue(currParam);
                     } else if(currParam[0] === 'environment' && pcEnv === null) {
                         pcEnv = getParamValue(currParam);
+                    } else if(currParam[0] === 'conversationId') {
```
justinaray, line 87: I'm assuming my thought process was to use a different queryParamName than the replacement token to show the variable could be different.  But, in hind sight, maybe that's confusing?

Thoughts about supporting both param names for pcLangTag and pcConversationId?  For instance:
`['langTag', 'pcLangTag'].indexOf(paramName) === 0` and
`['conversationId', 'pcConversationId'].indexOf(paramName) === 0`?

carolinepar: Yeah I see what you mean. That's a good idea, I'll add that.

carolinepar: Your syntax still only includes the first index, should it be `!== -1` instead of `=== 0`?

examples/urlInterpolationDemo.html
```diff
@@ -41,6 +41,7 @@ <h2>Localization Demo</h2>
             (function () {
                 var DEFAULT_LANG_TAG = 'en-us';
                 var langTag = null;
+                var conversationId = null;
```
justinaray, line 44: nit: this conversationId var is not needed at this scope as it's unused in the rest of the function.

justinaray (approved):

justinaray (approved):

PR #90, opened 2020-12-09
ATEAM-321: Target valid platform client environments in examples

jakejrichards:
When the host app is running locally, we need a way to specify which environment the platform client should target. This reads from a new env variable `CLIENT_APP_SDK_PC_DEV_PLATFORM_ENV` and is injected as part of the example transformations.

### DIFF
--- a/examples/utils/oauth.js
+++ b/examples/utils/oauth.js
@@ -1,15 +1,25 @@
 // Authenticate with Genesys Cloud
 function authenticate(client, pcEnvironment) {
+    // Allow targeting a different environment when host app is running locally
+    const platformEnvironment = pcEnvironment === 'localhost' ? 'mypurecloud.com' : pcEnvironment;
     /*
     * Note: To use this app in your own org, you will need to create your own OAuth2 Client(s)
     * in your Genesys Cloud org.  After creating the Implicit grant client, map the client id(s) to
     * the specified region key(s) in the object below, deploy the page, and configure an app to point to that URL.
     */
     const pcOAuthClientIds = {'mypurecloud.com': 'implicit-oauth-client-id-here'};
-    const clientId = pcOAuthClientIds[pcEnvironment];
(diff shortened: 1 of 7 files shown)


examples/utils/oauth.js
```diff
@@ -1,16 +1,21 @@
...
     /*
     * Note: To use this app in your own org, you will need to create your own OAuth2 Client(s)
     * in your Genesys Cloud org.  After creating the Implicit grant client, map the client id(s) to
     * the specified region key(s) in the object below, deploy the page, and configure an app to point to that URL.
     */
     const pcOAuthClientIds = {'mypurecloud.com': 'implicit-oauth-client-id-here'};
-    const clientId = pcOAuthClientIds[pcEnvironment];
+    const clientId = pcOAuthClientIds[platformEnvironment];
     if (!clientId) {
         return Promise.reject(new Error(pcEnvironment + ': Unknown/Unsupported Genesys Cloud Environment'));
```
justinaray, line 13: nit: you could use platformEnvironment, maybe with an extra note when localhost, here so the dev knows which env localhost was mapped to and that it's not in their oauth client list.

scripts/build-examples.js
```diff
@@ -30,12 +33,22 @@ function transformSdkOAuthClientIds(buffer, ids) {
     );
 }
 
+function transformPlatformEnvironment(buffer, env) {
+    return buffer.replace(
+        /(platformEnvironment =)[^;]+;/,
+        `$1 pcEnvironment === 'localhost' ? '${env}' : pcEnvironment;`
```
justinaray, line 39: so much escape.  code in a template literal in a regex replace!

:thumbsup:

justinaray (approved):
One little nit on error logging but otherwise looks good.

Thanks for knocking this one out!

examples/utils/oauth.js
jakejrichards: ahh good catch 👍

PR #92, opened 2021-02-01
ATEAM-273: Add support for internal sdk extensions and pull in service-discovery

jakejrichards:
Modify the signature of lookupEnv so that we can reuse and override it in the internal SDK class and pull in service discovery

### DIFF
--- a/src/utils/envSpec.ts
+++ b/src/utils/envSpec.ts
@@ -1,34 +1,39 @@
 import * as envUtils from './env';
 
 export default describe('env utils', () => {
-    const VALID_PC_TLDS = [
-        'mypurecloud.com',
-        'mypurecloud.com.au',
-        'mypurecloud.ie',
-        'mypurecloud.jp',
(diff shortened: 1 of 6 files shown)


src/utils/envSpec.ts
```diff
@@ -1,4 +1,4 @@
-import envUtils from './env';
+import * as envUtils from './env';
 
 export default describe('env utils', () => {
     const VALID_PC_TLDS = [
```
justinaray, line 4: I could go either way here.  Keeping this list up-to-date or just having tests for the unique domain patterns we have here so we get out of the business of tracking this list.  Thoughts?

jakejrichards: Ah yeah, didn't realize we had tests around the tlds. I think I'm in favor of killing these and not tracking the list.

justinaray (approved):
Looks good to me @jakejrichards.  One comment on unit test but not strictly necessary to change.

PR #106, opened 2021-07-26
Fixed external contact error in agent-interactions-demo

poweitsao:
In Agent Interactions Demo, if the interaction is not chat (e.g. outbound email) and the contact is an external contact, an extra API call to ExternalContactsAPI is needed to get customer name.


examples/agentInteractionsDemo/app.js
```diff
@@ -207,9 +209,17 @@ new Vue({
                                 .then((conv) => {
                                     let newConv = conv;
                                     newConv.customer = conv.participants.find((part) => {
-                                        return part.purpose === "customer";
+                                        return part.purpose === "customer" || part.purpose === "external";
                                     });
-                                    this.conversationsData.conversations.push(newConv);
+                                    if (newConv.customer.purpose === "external"){
```
jakejrichards, line 214: Technically, I think this logic would be best as `if (newConv.customer.externalContactId) {...}` since we could still have a linked contact with `purpose !== 'external'` - we'd just get the "more correct" name in this case.

examples/agentInteractionsDemo/app.js
```diff
@@ -207,9 +209,17 @@ new Vue({
                                 .then((conv) => {
                                     let newConv = conv;
                                     newConv.customer = conv.participants.find((part) => {
```
jakejrichards, line 211: In the scenario where this `find` doesn't yield a customer (not sure if possible), should we be a bit more robust and fallback to `{ name: 'Unknown' }` so that the conversation can still render?

Perhaps something like:
```
newConv.customer = conv.participants.find(...) || { name: 'Unknown" };
```

examples/agentInteractionsDemo/app.js
poweitsao: Ah, I totally missed this edge case. Thanks!

examples/agentInteractionsDemo/app.js
poweitsao: Good idea! It would definitely make it the demo more robust.

examples/agentInteractionsDemo/app.js
```diff
@@ -207,9 +209,22 @@ new Vue({
...
+                                    if (newConv.customer.externalContactId){
+                                        console.log("externalContact", newConv)
+                                        externalContactsApi.getExternalcontactsContact(newConv.customer.externalContactId)
+                                            .then((externalContact) => {
+                                                newConv.customer.name = `${externalContact.firstName} ${externalContact.lastName}`;
+                                                this.conversationsData.conversations.push(newConv);
+                                            })
+                                    } else{
+                                        if(!newConv.customer.name){
+                                            newConv.customer.name = "Unknown";
```
poweitsao, line 224: Added this here because there's an edge case where an agent sends an outbounding email to someone that is not an external contact, the conversation object will not have the name field.

examples/agentInteractionsDemo/app.js
```diff
@@ -175,9 +175,11 @@ new Vue({
         const usersApi = new platformClient.UsersApi();
         const qualityApi = new platformClient.QualityApi();
         const conversationsApi = new platformClient.ConversationsApi();
+        const externalContactsApi = new platformClient.ExternalContactsApi();
         Vue.prototype.$usersApi = usersApi;
         Vue.prototype.$qualityApi = qualityApi;
         Vue.prototype.$conversationsApi = conversationsApi;
+        Vue.prototype.$externalContactsApi = externalContactsApi;
```
justinaray, line 182: Does anyone know if you also need to initialize the Vue.prototype.$externalContactsApi to null at the top of the file?

I dont see any direct refs to any of the $ vars on the Vue instance anyway, so, from a js perspective this seems unnecessary, but, not sure of the vueisms that might need this.

examples/agentInteractionsDemo/app.js
```diff
@@ -207,9 +209,22 @@ new Vue({
...
                                     let newConv = conv;
                                     newConv.customer = conv.participants.find((part) => {
-                                        return part.purpose === "customer";
-                                    });
-                                    this.conversationsData.conversations.push(newConv);
+                                        return part.purpose === "customer" || part.purpose === "external";
+                                    }) || { name: "Unknown" };
+
+                                    if (newConv.customer.externalContactId){
+                                        console.log("externalContact", newConv)
```
justinaray, line 216: might want to remove this left over log statement if it was only for debugging.

examples/agentInteractionsDemo/app.js
```diff
@@ -207,9 +209,22 @@ new Vue({
                                 .then((conv) => {
                                     let newConv = conv;
                                     newConv.customer = conv.participants.find((part) => {
-                                        return part.purpose === "customer";
-                                    });
-                                    this.conversationsData.conversations.push(newConv);
+                                        return part.purpose === "customer" || part.purpose === "external";
```
justinaray, line 212: Since you're using this in two places, would it make sense to create a util method, getCustomerParticipant(conversation) that encapsulates this logic?  If it returns null, you can still handle the errors here where the call is made.

This will also let you improve this logic a bit.  For instance, you might want to return the last customer or external participant vs the first.  There are some other nuanced cases for detecting the customer, but, it's probably overkill for this demo.

examples/agentInteractionsDemo/app.js
```diff
@@ -207,9 +209,22 @@ new Vue({
...
+                                    }) || { name: "Unknown" };
+
+                                    if (newConv.customer.externalContactId){
+                                        console.log("externalContact", newConv)
+                                        externalContactsApi.getExternalcontactsContact(newConv.customer.externalContactId)
+                                            .then((externalContact) => {
+                                                newConv.customer.name = `${externalContact.firstName} ${externalContact.lastName}`;
+                                                this.conversationsData.conversations.push(newConv);
```
justinaray, line 220: below, on 228, the code attempts to set a unique array and also builds up a map of convo id to evals.  I think the eval map will still work with this additional promise and not waiting on it, but, the unique array might be a bit wonky.

In an ideal world, I think we make an async util call that resolves with the convos and evals and then just assign it to vue's model.  That way, we're not constantly repainting when each conversation details resolve.  But, you could also just extract the uniqueness and map building into a function and call that in two places if you dont want to get into all this for a bugfix.  Dealer's choice.

examples/agentInteractionsDemo/app.js
```diff
@@ -175,13 +175,53 @@ new Vue({
...
+        const externalContactsApi = new platformClient.ExternalContactsApi();
         Vue.prototype.$usersApi = usersApi;
         Vue.prototype.$qualityApi = qualityApi;
         Vue.prototype.$conversationsApi = conversationsApi;
+        Vue.prototype.$externalContactsApi = externalContactsApi;
 
         let authenticated = false;
         let agentUserId = "";
 
+        function getCustomerParticipant(conv) {
```
jakejrichards, line 187: So it's kind of a nit since this is a demo app, but I'll throw it out there anyway. Generally, we like to keep functions pure, highlighting any mutation or side effects with an explicit function name. It can quickly become unclear where mutations occur if we don't highlight them.

In this case, the conversation is passed in and we're mutating the attached customer participant. I think we should either a) change the function name, b) isolate the mutation and make it more obvious, or c) avoid the mutation altogether.

In an ideal world, I think we'd avoid this mutation altogether, but given that we were already doing this I think `b` is the best. I think a better way to delegate the responsibilities here could be something like:

```
const customer = getCustomerParticipant(conversation);
customer.name = await getCustomerName(customer);
```

^ Then it's left up to the consumers to decide whether or not the conversation should be mutated.

examples/agentInteractionsDemo/app.js
```diff
@@ -175,13 +175,53 @@ new Vue({
...
+                if(!newConv.customer.name){
+                    newConv.customer.name = "Unknown";
+                }
+            } 
+            return Promise.resolve(newConv)
+        }
+
+        async function getEvalConversationsData(evaluations){
+            const promises = [];
+            evaluations.forEach((eval) => {
```
jakejrichards, line 209: I think this is a perfect use case for `.map`:
```
return await Promise.all(
    evaluations.map(eval => conversationsApi.getConversation(eval.conversation.id).then(...).catch(...))
);
```

examples/agentInteractionsDemo/app.js
```diff
@@ -175,13 +175,56 @@ new Vue({
...
+
+        async function getEvalConversationsData(evaluations){
+
+            const evalConversations = await Promise.all(
+                evaluations.map(eval => conversationsApi.getConversation(eval.conversation.id).then(
+                    async (conv) => {
+                        conv.customer = getCustomerParticipant(conv);
+                        const name = await getCustomerName(conv.customer);
+                        conv.customer.name = name;
+                        return await Promise.resolve(conv);
```
justinaray, line 219: you can just `return conv` here, right?

examples/agentInteractionsDemo/app.js
```diff
@@ -175,13 +175,56 @@ new Vue({
...
+                    }
+                ).catch((err) => {
+                    console.log(`Error: ${err}`);
+                }))
+            );
+            return await Promise.resolve(evalConversations);
```
justinaray, line 225: Same here, since you're in an async function, you can just return and that will resolve the promise.

examples/agentInteractionsDemo/app.js
```diff
@@ -202,25 +245,18 @@ new Vue({
...
-                                .then((conv) => {
-                                    let newConv = conv;
-                                    newConv.customer = conv.participants.find((part) => {
-                                        return part.purpose === "customer";
-                                    });
-                                    this.conversationsData.conversations.push(newConv);
+                        getEvalConversationsData(evaluations)
+                            .then((evalConversations) => {
+                                evalConversations.forEach((evalConv, index) => {
+                                    this.conversationsData.conversations.push(evalConv);
```
justinaray, line 251: Since `getEvalConversationsData` catches errors and returns `undefined`, you'll want to filter out falsey evalConv instances.  It's important to just return here as the index is used to match the conversation with the evaluations.

alternatively, we could rework this a bit and make one function that combines both getConversations and getEvaluations and results in a data structure that is an array of
{
convo: Conversation,
evals: Evaluation[]
}

This would reduce the need to constantly check for uniqueness in the conversations array and it would be a bit more clear when we're in a data fetch state.  Granted, the conversations won't come in as they're resolved, but I'm not sure that's necessary.

This is just a bug fix to an existing impl tho, so I'm good either way.

examples/agentInteractionsDemo/app.js
```diff
@@ -175,13 +175,88 @@ new Vue({
...
+
+        async function getConversationsAndEvaluations(agentUserId) {
+            const startTime = moment('1997-01-01').toISOString();
+            const endTime = moment().toISOString();
+            return getEvaluations(qualityApi, startTime, endTime, agentUserId)
+                .then(async (data) => {
+                    const evaluations = data.entities;
+                    const evalConversations = {}
+
+                    return Promise.all(
```
justinaray, line 232: since you're in an async fn, should we change this to await?

examples/agentInteractionsDemo/app.js
```diff
@@ -194,53 +269,12 @@ new Vue({
...
-                    })
-                    .catch((err) => {
-                        console.log(`Error: ${err}`);
-                    });
+                getConversationsAndEvaluations(agentUserId).then(conversations => {
+                    for(var convId in conversations){
+                        this.conversationsData.conversations.push(conversations[convId].conv)
+                        if(conversations[convId].evals.length > 0) this.conversationsData.convEvalMap.set(convId, conversations[convId].evals);
+                    }
+                })
```
justinaray, line 277: might be worth adding a catch block here and setting `this.errorMessage` to `Failed to fectch conversations/evaluations`.

examples/agentInteractionsDemo/app.js
justinaray: There's actually a handful of places where we could use async/await.  I'm not 100% sure on the promise impl of the sdk.  So, if these are necessary, feel free to ignore as async/await is just style/readability.

examples/agentInteractionsDemo/app.js
```diff
@@ -175,13 +175,88 @@ new Vue({
...
+
+                }).then(async (conversations) => {
+                    return conversationsApi.getConversations().then(async (data)=> {
+                        await Promise.all(
+                            data.entities.filter(conv => !(conv.id in conversations))
+                                .map(conv_1 => {
+                                    conversations[conv_1.id] = {
+                                        evals: []
+                                    };
+                                    return getConversationData(conv_1.id).then(newConv => conversations[newConv.id].conv = newConv);
```
justinaray, line 252: where you seeing cases where the conversation data payload differed from getConversations and getConversation?

examples/agentInteractionsDemo/app.js
```diff
@@ -175,13 +175,88 @@ new Vue({
...
+                    const evalConversations = {}
+
+                    return Promise.all(
+                        evaluations.map(async eval => {
+                            const evalConvId = eval.conversation.id
+                            if (!(evalConvId in evalConversations)){
+                                evalConversations[evalConvId] = {evals: []}
+                            }
+                            evalConversations[evalConvId].evals.push(eval)
+                            const evalConv = await getConversationData(evalConvId);
```
justinaray, line 239: looks like getConversationData catches, logs, and returns undefined.

you have a few options imo,
* check for undefined and don't add the eval or convo to the list
* let it throw, change `Promise.all` to `Promise.allSettled` and then filter out evals with invalid convos
* allow the convo to be undefined and handle it in the template.

I'm basing that

examples/agentInteractionsDemo/app.js
```diff
@@ -175,13 +175,88 @@ new Vue({
...
         Vue.prototype.$conversationsApi = conversationsApi;
+        Vue.prototype.$externalContactsApi = externalContactsApi;
 
         let authenticated = false;
         let agentUserId = "";
 
+        function getCustomerParticipant(conv) {
+            let newConv = conv;
+            newConv.customer = newConv.participants.find((part) => {
+                return part.purpose === "customer" || part.purpose === "external";
```
justinaray, line 190: should we also verify that the participant has a name prop here?

justinaray: ah, nevermind.  I see the check below now.

But, a good validation of Jake's comment to make the function functional or rename to make the mutation obvious.

Maybe return the customerParticipant or null, pass that into getCustomerName, and then build up the resulting customer object and add it in `getConversationData`?

justinaray:
@poweitsao ... looks good and fixes the bug!  Added some stylistic and nitty comments but I think the general impl is much improved.

Cool to be able to use promise, async/await, map, etc. natively now!

examples/agentInteractionsDemo/app.js
poweitsao: I should have named my variables better, sorry. Basically, this Promise.all gets all the conversations while the previous one only gets the ones with evaluations. So I'm just filtering out the conversations that we already got from the previous Promise.all. Is that what your question was asking about?

justinaray (approved):

jakejrichards (approved):

PR #115, opened 2021-11-17
Feature/ateam 601 add show group sdk method

carolinepar:
If we're happy with moving show profile to the ui module I can delete ui, let me know if it's good to remove


justinaray:
Replaced by https://github.com/MyPureCloud/client-app-sdk/pull/117

PR #117, opened 2021-12-15
Feature/ateam 601 show group sdk method and example

justinaray:


examples/directoryNavigationDemo.html
```diff
@@ -150,6 +169,33 @@
                 setHidden(failureEl, false);
             }
 
+            /**
+             * Updates the Group List element state with the user's group data
+             */
+            function setGroupList(listContainer, groupIds) {
+                listContainer.textContent = '';
+                Promise.all(groupIds.map(currGroupId => new platformClient.GroupsApi().getGroup(currGroupId.id)))
```
justinaray, line 177: will move instantiation out of the map

examples/directoryNavigationDemo.html
```diff
@@ -150,6 +169,33 @@
...
+                    });
+            }
+
+            const groupItemTemplate = document.getElementById('group-item-template');
+            function buildGroupItemEl(group) {
+                const result = groupItemTemplate.content.cloneNode(true);
+
+                const groupLinkEl = result.querySelector("a");
+                groupLinkEl.id = group.id;
+                groupLinkEl.appendChild(document.createTextNode(group.name));
```
justinaray, line 190: will use id as fallback in case name is empty

src/index.ts
```diff
@@ -83,9 +84,25 @@ class ClientApp {
      *
      * clientApp.users.someMethod(...);
      * ```
+     *
+     * @deprecated Since 2.3.0.  @see [directory](#directory) for replacements
```
justinaray, line 88: We decided to deprecate the users module and api and replace it with the directory api since it better describes/encapsulates the sdk methods of the module.  We'll remove this in 3.0.0

carolinepar (approved):

PR #127, opened 2022-03-10
fix(examples proposeMessage): fix target handling in example event li…

justinaray:
…steners

Switch to currentTarget to get the event listener node vs the clicked
node.  This caused nested HTML inside the demo string buttons to get
sent versus the innerHTML of the entire button.

### DIFF
--- a/examples/proposeMessageDemo.html
+++ b/examples/proposeMessageDemo.html
@@ -84,7 +84,7 @@
         }
 
         customResponse.addEventListener('submit', event => {
-            let message = event.target[0].value;
+            let message = event.currentTarget[0].value;
             console.log('form: ', message);
             clientApp.conversations.proposeInteractionMessage('insert', message);
             event.preventDefault();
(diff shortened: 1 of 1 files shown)


examples/proposeMessageDemo.html
```diff
@@ -84,15 +84,15 @@
...
+            let message = event.currentTarget[0].value;
             console.log('form: ', message);
             clientApp.conversations.proposeInteractionMessage('insert', message);
             event.preventDefault();
         });
 
         precannedButtons.forEach(button =>
             button.addEventListener('click', event => {
-                let message = event.target.innerHTML;
+                let message = event.currentTarget.innerHTML;
```
justinaray, line 95: target is the clicked element while currentTarget is the element to which the listener was attached.  We were getting nested html rather than the full innerHTML of the button.

carolinepar (approved):

carolinepar (approved):

PR #128, opened 2022-03-14
Feature/ateam 632 push the host origin down

bpoc:
Adds the Genesys cloud host origin and target environment as variables. If gcHostOrigin is set then it will use that for the window.postMessage target origin

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "purecloud-client-app-sdk",
-  "version": "2.3.0",
+  "version": "2.3.1",
   "description": "Javascript API for integrating 3rd-party apps with Genesys Cloud",
   "main": "dist/main.js",
   "jsnext:main": "dist/main.mjs",
--- a/src/index.ts
+++ b/src/index.ts
@@ -36,6 +36,17 @@ class ClientApp {
     private _customPcOrigin: string | null = null;
 
     /**
+     * The private reference to the Genesys Cloud Host Origin, if provided.
+     */
+    private _gcHostOrigin: string | null = null;
+
+    /**
+     * Variable for the name of the type of environment
+     */
+    // @ts-ignore
+    private gcTargetEnv: string | null = null; 
+
+    /**
      * The AlertingApi instance.
      *
      * ```ts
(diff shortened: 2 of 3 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "purecloud-client-app-sdk",
-  "version": "2.3.0",
+  "version": "2.3.1",
```
justinaray, line 3: build can bump this so shouldn't need to increment in the PR.

src/index.ts
```diff
@@ -158,6 +169,10 @@ class ClientApp {
      * @param cfg - Runtime config of the client
      */
     constructor(cfg: {
+        /** The origin of the genesys cloud parent. Overrides the pcEnvironment and functions similar to pcOrigin except the gcHostOrigin can be passed by the genesys cloud parent*/
+        gcHostOriginQueryParam?: string;
+        /** The type of environment the genesys cloud parent is running under(e.g. dev, test, prod)*/
+        gcTargetEnvQueryParam?: string;
```
justinaray, line 175: maybe for convenience we support these as either queryParams or strings with the string values taking precedence if both happen to be specified.

The queryParam is how we expect most apps to configure and get their config so we provide these as convenience methods.  However, there are cases where an app might need to parse the values from elsewhere or just manually pass in

src/index.ts
```diff
@@ -180,15 +202,24 @@ class ClientApp {
...
                 this._customPcOrigin = cfg.pcOrigin;
             }
+
+            if("gcTargetEnvQueryParam" in cfg){
+                const paramName = cfg.gcTargetEnvQueryParam;
+                this.assertNonEmptyString(paramName, 'target env query param name');
+                const parsedQueryString = queryString.parse(ClientApp._getQueryString() || '');
+                const paramValue = decodeURI(parsedQueryString[paramName]);
+                this.gcTargetEnv = paramValue;
+            }
```
justinaray, line 212: ultimately, i think we need to parse/acquire this soup of values and then build up a PcEnv representation.  We might want to add environment to that type in addition to the hostname.  We should be able to map envs to tlds via sd-web.  If we consistently set the internal `_pcEnv` prop, then the `pcEnvironment` will still be supported.

The other option is that we could rev the major version and just support host and target env.  Let's chat about this tomorrow.

PR #146, opened 2023-01-11
Consume host origin and target env

cadgerfeast:
Here is a first draft for the host origin and target env support.
This one should be non-breaking, but this implies some duplicated stuff, more info below.
Here is a short description of what has been added:

- `.env` file support with [dotenv](https://github.com/motdotla/dotenv)
- New `CLIENT_APP_SDK_GC_EXTRA_ENVS` env variable, that can contain an array of environments
- Various fixes for windows development
- `gcHostOriginQueryParam` and `gcTargetEnvQueryParam` params to construct client app support (must be used together)
- a few unit-tests

About the new `CLIENT_APP_SDK_GC_EXTRA_ENVS`, I think we should remove the existing `CLIENT_APP_SDK_PC_DEV_ENVS` variables, and compute the values from the new one, what do you think?
Also, I only updated the `toast.html` sample, should I do the same for all other samples?

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -85,6 +85,14 @@ Whitelist additional comma-delimited top level domains (e.g. example.com).
 
 #### CLIENT_APP_SDK_GC_EXTRA_ENVS
 Adds a list of Genesys Cloud environments that should be supported.
+``` json
+[
+    {
+        "env": "dev",
+        "publicDomainName": "some-domain-name.ext"
+    }
+]
+```
 
 #### CLIENT_APP_SDK_PC_OAUTH_CLIENT_IDS
 Specify a JSON string mapping environment to OAuth client id to be injected into example html files.
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -83,6 +83,9 @@ Specify the Genesys Cloud environment to target when host app is running on `loc
 #### CLIENT_APP_SDK_PC_DEV_ENVS
 Whitelist additional comma-delimited top level domains (e.g. example.com).
 
+#### CLIENT_APP_SDK_GC_EXTRA_ENVS
+Adds a list of Genesys Cloud environments that should be supported.
+
 #### CLIENT_APP_SDK_PC_OAUTH_CLIENT_IDS
 Specify a JSON string mapping environment to OAuth client id to be injected into example html files.
 ```json
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -89,7 +89,7 @@ Adds a list of Genesys Cloud environments that should be supported.
 [
     {
         "env": "dev",
-        "publicDomainName": "some-domain-name.ext"
+        "publicDomainName": "some-domain-name.tld"
     }
 ]
 ```
--- a/dev.rollup.config.js
+++ b/dev.rollup.config.js
@@ -6,8 +6,10 @@ import path from 'path';
 import serve from 'rollup-plugin-serve';
 import livereload from 'rollup-plugin-livereload';
 import { umdConfig } from './rollup.config';
+
 const { buildExample, buildExamples } = require('./scripts/build-examples');
 
+const examplesPath = path.resolve('examples');
 const tmpDestPath = fs.mkdtempSync(path.join(os.tmpdir(), `${pkg.name}-dev-server-`));
 
 let httpsConfig;
@@ -21,14 +23,15 @@ try {
 }
 
 const tsc = () => {
-    return new Promise((resolve) => {
-        const bin = require.resolve('typescript/bin/tsc');
+    return new Promise((resolve, reject) => {
+        const bin = process.platform === 'win32' ? require.resolve('.bin/tsc.cmd') : require.resolve('.bin/tsc');
         spawn(bin, [
             '--incremental',
             '--outDir', `${tmpDestPath}/ts-build`,
             '--project', 'tsconfig.build.json'
         ], { stdio: 'inherit' })
-            .on('exit', resolve);
+            .on('exit', resolve)
+            .on('error', (err) => reject(err));
     });
 };
 
@@ -40,20 +43,21 @@ export default Object.assign({}, umdConfig, {
         dir: tmpDestPath
     }),
     watch: {
-        clearScreen: false
+        clearScreen: false,
     },
     plugins: [
         ...umdConfig.plugins,
         {
             name: 'process-examples',
             async buildStart() {
-                this.addWatchFile(path.resolve('examples'));
+                this.addWatchFile(examplesPath);
                 await tsc(); // Check typescript types
             },
             watchChange(id) {
                 // Rebuild individual example file when changed
-                if (id.indexOf('examples/') < 0) return;
-                buildExample(tmpDestPath, path.relative(__dirname, id));
+                if (id.startsWith(examplesPath)) {
+                    buildExample(tmpDestPath, path.relative(__dirname, id));
+                }
             }
         },
         serve({
--- a/dev.rollup.config.js
+++ b/dev.rollup.config.js
@@ -43,7 +43,7 @@ export default Object.assign({}, umdConfig, {
         dir: tmpDestPath
     }),
     watch: {
-        clearScreen: false,
+        clearScreen: false
     },
     plugins: [
         ...umdConfig.plugins,
--- a/examples/toast.html
+++ b/examples/toast.html
@@ -160,10 +160,20 @@
                 // Note: This manual check for query string is for backwards compatibility of this
                 // deployed example.  In your own apps, you can assume the query param will be
                 // provided by Genesys Cloud if you have configured it in your app's config.
-                let envQueryParamName = 'pcEnvironment';
-                if (window && window.location && typeof window.location.search === 'string' &&
-                    window.location.search.indexOf(envQueryParamName) >= 0) {
-                    myClientApp = new window.purecloud.apps.ClientApp({pcEnvironmentQueryParam: envQueryParamName});
+                const envQueryParamName = 'pcEnvironment';
+                const hostQueryParamName = 'gcHostOrigin';
+                const targetEnvQueryParamName = 'gcTargetEnv';
+                const locationSearch = (window && window.location && typeof window.location.search === 'string') ? window.location.search : '';
+                const queryParams = new URLSearchParams(locationSearch);
+                if (queryParams.get(hostQueryParamName) && queryParams.get(targetEnvQueryParamName)) {
+                    // Compute Genesys Cloud region from host origin
+                    myClientApp = new window.purecloud.apps.ClientApp({
+                        gcHostOriginQueryParam: hostQueryParamName,
+                        gcTargetEnvQueryParam: targetEnvQueryParamName
+                    });
+                } else if (queryParams.get(envQueryParamName)) {
+                    // Compute Genesys Cloud region from pcEnvironment
+                    myClientApp = new window.purecloud.apps.ClientApp({ pcEnvironmentQueryParam: envQueryParamName });
                 } else {
                     // Use default Genesys Cloud region
                     myClientApp = new window.purecloud.apps.ClientApp();
--- a/examples/toast.html
+++ b/examples/toast.html
@@ -165,7 +165,7 @@
                 const targetEnvQueryParamName = 'gcTargetEnv';
                 const locationSearch = (window && window.location && typeof window.location.search === 'string') ? window.location.search : '';
                 const queryParams = new URLSearchParams(locationSearch);
-                if (queryParams.get(hostQueryParamName) && queryParams.get(targetEnvQueryParamName)) {
+                if (queryParams.get(hostQueryParamName) || queryParams.get(targetEnvQueryParamName)) {
                     // Compute Genesys Cloud region from host origin
                     myClientApp = new window.purecloud.apps.ClientApp({
                         gcHostOriginQueryParam: hostQueryParamName,
--- a/package.json
+++ b/package.json
@@ -86,6 +86,7 @@
     "@typescript-eslint/eslint-plugin": "^3.2.0",
     "@typescript-eslint/parser": "^3.2.0",
     "child-process-promise": "^2.0.2",
+    "dotenv": "^16.0.3",
     "eslint": "^7.2.0",
     "eslint-config-standard": "^11.0.0-beta.0",
     "eslint-plugin-import": "^2.8.0",
--- a/scripts/build-examples.js
+++ b/scripts/build-examples.js
@@ -9,7 +9,6 @@ const {
     CLIENT_APP_SDK_PC_DEV_PLATFORM_ENV: devPlatformEnv
 } = process.env;
 const BROWSER_FILENAME = `/${pkg.name}.js`;
-const examplesPath = path.resolve('examples');
 
 // Called from command line
 if (!module.parent) {
@@ -60,9 +59,8 @@ function buildExample(outDir, relativeFilePath, bundleFileName) {
             buffer = transformPlatformEnvironment(buffer, devPlatformEnv);
         }
     }
-
     fs.outputFileSync(
-        path.join(outDir, path.relative(examplesPath, relativeFilePath)),
+        path.join(outDir, relativeFilePath.replace(/examples[/|\\]/, '')),
         buffer
     );
 }
--- a/scripts/build-examples.js
+++ b/scripts/build-examples.js
@@ -9,6 +9,7 @@ const {
     CLIENT_APP_SDK_PC_DEV_PLATFORM_ENV: devPlatformEnv
 } = process.env;
 const BROWSER_FILENAME = `/${pkg.name}.js`;
+const examplesPath = path.resolve('examples');
 
 // Called from command line
 if (!module.parent) {
@@ -61,7 +62,7 @@ function buildExample(outDir, relativeFilePath, bundleFileName) {
     }
 
     fs.outputFileSync(
-        path.join(outDir, relativeFilePath.replace('examples/', '')),
+        path.join(outDir, path.relative(examplesPath, relativeFilePath)),
         buffer
     );
 }
@@ -69,7 +70,7 @@ function buildExample(outDir, relativeFilePath, bundleFileName) {
 function buildExamples(outDir, bundleFileName) {
     glob
         .sync('examples/**/*', { nodir: true })
-        .forEach(example => buildExample(outDir, example, bundleFileName));
+        .forEach((example) => buildExample(outDir, path.normalize(example), bundleFileName));
 }
 
 module.exports = { buildExample, buildExamples };
--- a/src/ClientAppSpec.ts
+++ b/src/ClientAppSpec.ts
@@ -183,7 +183,7 @@ export default describe('ClientApp', () => {
                         gcTargetEnvQueryParam: 'gcTargetEnv'
                     });
                 }).toThrow();
-                // Localhost
+                // Invalid target env 
                 query = '?gcHostOrigin=https://apps.mypurecloud.com&gcTargetEnv=invalid';
                 expect(() => {
                     new ClientApp({
--- a/src/ClientAppSpec.ts
+++ b/src/ClientAppSpec.ts
@@ -169,7 +169,6 @@ export default describe('ClientApp', () => {
                 }).toThrow();
                 // External
                 query = '?gcHostOrigin=https://invalid.com&gcTargetEnv=prod';
-                spyOn(ClientApp, '_getQueryString').and.callFake(() => query);
                 expect(() => {
                     new ClientApp({
                         gcHostOriginQueryParam: 'gcHostOrigin',
--- a/src/ClientAppSpec.ts
+++ b/src/ClientAppSpec.ts
@@ -158,8 +158,17 @@ export default describe('ClientApp', () => {
                 expect(myClientApp.pcEnvironment).toBe('localhost');
             });
             it('should fail if is not targeting a valid env', () => {
+                // Mismatch
+                let query = '?gcHostOrigin=https://apps.mypurecloud.jp&gcTargetEnv=prod';
+                spyOn(ClientApp, '_getQueryString').and.callFake(() => query);
+                expect(() => {
+                    new ClientApp({
+                        gcHostOriginQueryParam: 'gcHostOrigin',
+                        gcTargetEnvQueryParam: 'gcTargetEnv'
+                    });
+                }).toThrow();
                 // External
-                let query = '?gcHostOrigin=https://invalid.com&gcTargetEnv=prod';
+                query = '?gcHostOrigin=https://invalid.com&gcTargetEnv=prod';
                 spyOn(ClientApp, '_getQueryString').and.callFake(() => query);
                 expect(() => {
                     new ClientApp({
--- a/src/ClientAppSpec.ts
+++ b/src/ClientAppSpec.ts
@@ -158,9 +158,17 @@ export default describe('ClientApp', () => {
                 expect(myClientApp.pcEnvironment).toBe('localhost');
             });
             it('should fail if is not targeting a valid env', () => {
-                // Mismatch
+                // Mismatch Environment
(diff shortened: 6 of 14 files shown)


src/globals.d.ts
```diff
@@ -1,4 +1,5 @@
 declare const __PACKAGE_NAME__: string;
 declare const __PACKAGE_VERSION__: string;
 declare const __PC_DEV_ENVS__: string[];
+declare const __GC_DEV_EXTRA_ENVS__: unknown[];
```
justinaray, line 4: Would be great if we could declare this as `Environment[]`.

I played with this a bit and it's a bit unfortunate that adding an import then registers this as a module; thereby requiring that we export these declarations and then import them into the consuming files.

[this SO article](https://stackoverflow.com/questions/45420448/how-to-import-external-type-into-global-d-ts-file) suggests that we might be able to use a dynamic import to get around this issue.

Maybe you can take a look and see if there's a solution that we can type this correctly and not import everywhere?

If this turns into a research project or overly inflates this PR, we can leave it as unknown.  but, work a small look I think.

CONTRIBUTING.md
```diff
@@ -83,6 +83,9 @@ Specify the Genesys Cloud environment to target when host app is running on `loc
 #### CLIENT_APP_SDK_PC_DEV_ENVS
 Whitelist additional comma-delimited top level domains (e.g. example.com).
 
+#### CLIENT_APP_SDK_GC_EXTRA_ENVS
+Adds a list of Genesys Cloud environments that should be supported.
```
justinaray, line 87: :thumbsup: ... i was thinking about this after our chat and was going to mention we might need this.  Would you mind describing the type a bit?  maybe just Environment[] would be enough so devs will know the format of this value?

dev.rollup.config.js
```diff
@@ -21,14 +23,15 @@ try {
 }
 
 const tsc = () => {
-    return new Promise((resolve) => {
-        const bin = require.resolve('typescript/bin/tsc');
+    return new Promise((resolve, reject) => {
+        const bin = process.platform === 'win32' ? require.resolve('.bin/tsc.cmd') : require.resolve('.bin/tsc');
```
justinaray, line 27: thanks for adding/fixing this

dev.rollup.config.js
```diff
@@ -40,20 +43,21 @@ export default Object.assign({}, umdConfig, {
         dir: tmpDestPath
     }),
     watch: {
-        clearScreen: false
+        clearScreen: false,
```
justinaray, line 46: we need to add prettier to this repo.  i usually omit trailing commas, but, that's completely a style thing and solved by prettier.

no need to change this back.  just making an observation

examples/toast.html
```diff
@@ -160,10 +160,20 @@ <h1>Open Window Handler</h1>
...
                 // deployed example.  In your own apps, you can assume the query param will be
                 // provided by Genesys Cloud if you have configured it in your app's config.
-                let envQueryParamName = 'pcEnvironment';
-                if (window && window.location && typeof window.location.search === 'string' &&
-                    window.location.search.indexOf(envQueryParamName) >= 0) {
-                    myClientApp = new window.purecloud.apps.ClientApp({pcEnvironmentQueryParam: envQueryParamName});
+                const envQueryParamName = 'pcEnvironment';
+                const hostQueryParamName = 'gcHostOrigin';
+                const targetEnvQueryParamName = 'gcTargetEnv';
+                const locationSearch = (window && window.location && typeof window.location.search === 'string') ? window.location.search : '';
```
justinaray, line 166: just a heads up/justification ...

a long time ago, when we started this repo, we still had to support IE 9/10.  So, we also tried to keep the examples es5 compat in source.

So, if you see lets or overly protective checks, that's the historical justification.  Def fine to use modern features and assume we're running on no-less than Chromium Edge these days.  Safari is our new IE though :joy:

If you're curious, here are our [supported browsers](https://help.mypurecloud.com/articles/genesys-cloud-requirements/)

examples/toast.html
```diff
@@ -160,10 +160,20 @@ <h1>Open Window Handler</h1>
...
+                const queryParams = new URLSearchParams(locationSearch);
+                if (queryParams.get(hostQueryParamName) && queryParams.get(targetEnvQueryParamName)) {
```
justinaray, line 168: we can probably assume/enforce that both of these will be provided in this example code as well.

could switch this to an `||`

package.json
```diff
@@ -86,6 +86,7 @@
     "@typescript-eslint/eslint-plugin": "^3.2.0",
     "@typescript-eslint/parser": "^3.2.0",
     "child-process-promise": "^2.0.2",
+    "dotenv": "^16.0.3",
```
justinaray, line 89: TIL about dotenv.  very cool!

dev.rollup.config.js
```diff
@@ -40,20 +43,21 @@ export default Object.assign({}, umdConfig, {
...
             async buildStart() {
-                this.addWatchFile(path.resolve('examples'));
+                this.addWatchFile(examplesPath);
                 await tsc(); // Check typescript types
             },
             watchChange(id) {
                 // Rebuild individual example file when changed
-                if (id.indexOf('examples/') < 0) return;
-                buildExample(tmpDestPath, path.relative(__dirname, id));
+                if (id.startsWith(examplesPath)) {
```
justinaray, line 58: i wasn't 100% sure how rollup would process and send the id.  i.e. I wasn't sure if the id could be a temp file versus the proper src location or a different root.

But, I tested it and the watch is definitely working and it's just the dev server if anything changes int he future, so this one looks good to me 👍

scripts/build-examples.js
```diff
@@ -61,15 +62,15 @@ function buildExample(outDir, relativeFilePath, bundleFileName) {
     }
 
     fs.outputFileSync(
-        path.join(outDir, relativeFilePath.replace('examples/', '')),
+        path.join(outDir, path.relative(examplesPath, relativeFilePath)),
```
justinaray, line 65: this one might be slightly risky.  there's no requirement that the relative path passed to this function is from our examples dir.  Though, in practice (both for dev server and building our examples for deployment/testing) this does seem to be the case.

src/index.ts
```diff
@@ -10,7 +10,8 @@
  */
 
 import * as queryString from 'query-string';
-import { lookupPcEnv, PcEnv, DEFAULT_PC_ENV } from './utils/env';
+import { Environment } from 'genesys-cloud-service-discovery-web';
```
justinaray, line 13: could this be a type import?

src/index.ts
```diff
@@ -164,12 +165,20 @@ class ClientApp {
         pcEnvironment?: string;
         /** The full origin (protocol, host, port) of the Genesys Cloud host environment (e.g. https://apps.mypurecloud.com).  Prefer using pcEnvironment[QueryParam] over this property. */
         pcOrigin?: string;
+        /** Name of a query param to auto parse into the gcHostOrigin; Must be valid and used with gcTargetEnvQueryParam. */
+        gcHostOriginQueryParam?: string;
+        /** Name of a query param to auto parse into the gcTargetEnv; Must be valid and used with gcHostOriginQueryParam. */
```
justinaray, line 170: awesome comment and desc!

src/index.ts
```diff
@@ -164,12 +165,20 @@ class ClientApp {
...
+        gcTargetEnvQueryParam?: string;
     } = {}) {
         if (cfg) {
-            if ('pcEnvironmentQueryParam' in cfg) {
+            const parsedQueryString = queryString.parse(ClientApp._getQueryString() || '');
+            if ('gcHostOriginQueryParam' in cfg || 'gcTargetEnvQueryParam' in cfg) {
+                this.assertNonEmptyString(cfg.gcHostOriginQueryParam, 'host origin query param name');
+                this.assertNonEmptyString(cfg.gcTargetEnvQueryParam, 'target env query param name');
```
justinaray, line 177: nice, thanks for asserting both of these.  We'll make it a requirement that these are always provided together.

src/utils/env.ts
```diff
@@ -21,11 +20,23 @@ const PC_ENV_TLDS = environments
         return tlds;
     }, [] as string[])
     .concat(__PC_DEV_ENVS__);
+const GC_DEV_EXTRA_ENVS = environments.concat(__GC_DEV_EXTRA_ENVS__ as Environment[]);
```
justinaray, line 23: if we can figure out a way to define the type in globals.d.ts, we won't need this type assertion here.

justinaray: maybe we name this `ALL_GC_ENVS`?  `GC_DEV_EXTRA_ENVS` reads like it's just the extra envs and through me off for a second.

src/utils/env.ts
```diff
@@ -65,3 +76,20 @@ export const lookupPcEnv = (pcEnvTld: string, lenient = false, envTlds: string[]
...
     return null;
 };
+
+/**
+ * Attempts to locate a PC environment corresponding to the provided url
+ * @param url A string representing the Genesys Cloud environment top-level domain to search for
+ * @returns A Genesys Cloud environment object if found; null otherwise.
+ */
+export const lookupGcEnv = (url: string, targetEnv: string, envList: Environment[] = GC_DEV_EXTRA_ENVS): PcEnv|null => {
+    const hostLocation = new URL(url);
```
justinaray, line 86: this could fail to parse.  For instance if url is not absolute.  Looks like we would technically be fine to throw the error all the way up, but, given we can already return null from this function and we present a user friendly message, maybe we should add a try/catch here and return null instead of throwing?

src/utils/env.ts
```diff
@@ -65,3 +76,20 @@ export const lookupPcEnv = (pcEnvTld: string, lenient = false, envTlds: string[]
...
+    if (['localhost', '127.0.0.1'].indexOf(hostLocation.hostname) !== -1) {
+        return {
+            pcEnvTld: 'localhost',
+            pcAppOrigin: url
```
justinaray, line 90: maybe use `url.origin` here?  that way, we're assured to only use the origin in case extra path/query/hash data is accidentally included on the url

src/utils/env.ts
```diff
@@ -21,11 +20,23 @@ const PC_ENV_TLDS = environments
         return tlds;
     }, [] as string[])
     .concat(__PC_DEV_ENVS__);
+const GC_DEV_EXTRA_ENVS = environments.concat(__GC_DEV_EXTRA_ENVS__ as Environment[]);
 
 const [defaultEnv] = environments.filter(env => env.region === DEFAULT_ENV_REGION);
 
 export const DEFAULT_PC_ENV = buildPcEnv(defaultEnv.publicDomainName);
 
+function findPcEnvironment(hostname: string, targetEnv: string, envList: Environment[]): PcEnv|null {
```
justinaray, line 29: if it weren't for the extra domains, this would be super simple.  we could do:

```typescript
// parse is exported from sd-web
const parsedEnv = parse(hostname, {});
if (!parsedEnv || parsedEnv.name !== targetEnv) {
  return null;
  }
  return {
    pcEnvTld: parsedEnv.publicDomainName,
    pcAppOrigin: hostname
    }
```

This would achieve a few things:
1. built-in hardened parsing of hostnames to envs from sd-web
1. asserts that the provided hostname is in fact associated witht he targetEnv
1. would use the provided hostname which is preferred over us having to assume the subdomain is `apps`.

justinaray: the wrinkle though is of course the custom envs.  `parse` provides no option to customize the list of envs to search on.  So, we're never going to get the custom envs in parse.

I guess one option would be to use the default `parse` method (maybe down in `lookupGcEnv`) and then fall back to a more lenient parse that only considers the custom envs?  It is for dev anyway.

Thoughts?  We can also chat through this tomorrow.

src/utils/env.ts
```diff
@@ -65,3 +76,20 @@ export const lookupPcEnv = (pcEnvTld: string, lenient = false, envTlds: string[]
...
 };
+
+/**
+ * Attempts to locate a PC environment corresponding to the provided url
+ * @param url A string representing the Genesys Cloud environment top-level domain to search for
+ * @returns A Genesys Cloud environment object if found; null otherwise.
+ */
+export const lookupGcEnv = (url: string, targetEnv: string, envList: Environment[] = GC_DEV_EXTRA_ENVS): PcEnv|null => {
+    const hostLocation = new URL(url);
+    if (['localhost', '127.0.0.1'].indexOf(hostLocation.hostname) !== -1) {
```
justinaray, line 87: nit: you could use `includes` instead of `indexOf` here.  I know we're not consistent with that across the repo.  see previous comment about this project starting a long time ago with old browser support :joy:

justinaray:
looking great so far @cadgerfeast!

There are a few comments here, but, mostly nits and discussion points.  The only thing we really need to sort is how to parse envs and support the custom ones.

Cheers!

dev.rollup.config.js
cadgerfeast: I also prefer to omit them, I guess I forgot that one during development ;)

scripts/build-examples.js
cadgerfeast: I understand, will change that.
The reason for this modification was with `windows` development support, `glob` sends a `relativePath` with `/` even when running on a `windows` machine, but it's not the case for `path` module.
We can keep the `replace()` method but with support replacing backslash characters.

src/utils/env.ts
cadgerfeast: One other solution might be to add an extra `extra` boolean property for dev envs (added when using `.concat(`) .
That way we could use the parse method, or this custom one for extra envs.

scripts/build-examples.js
justinaray: 👍 ... assumed it was for windows dev and thanks for adding support.

And, thanks for switching to the regexp approach.  probably safest fix.

src/globals.d.ts
justinaray: nice!  great to have the type declared here!

src/utils/env.ts
justinaray: i do like the `GCEnvironment` and `extra` approach, but now, i'm realizing we don't even need to merge in envs and the __GC_DEV_EXTRA_ENVS__ into ALL_GC_ENVS.

`parse` will internally loop through it's known envs searching for matches.  So, we can call parse first and if it doesn't find an env, we can fall back to searching through the extra envs.  And, I think we can just change the `findPcEnvironment` method signature to just take a list of the extra envs, negating the need for the extra prop.  It's also a bit cleaner.

Final point, instead of trying to match `apps.${p}`, we should switch it to location.hostname.endsWith(p).  This will allow us to match any subdomain (i.e. apps or app).  Then, we can return

```
return {
pcEnvTld: environment.publicDomainName,
pcAppOrigin: location.origin
}
```

This gives us full support of localhost, extra envs, and match any subdomain (apps, app, etc) 🤘

src/ClientAppSpec.ts
```diff
@@ -139,6 +139,90 @@ export default describe('ClientApp', () => {
...
+                    gcTargetEnvQueryParam: 'gcTargetEnv'
+                });
+                expect(myClientApp.pcEnvironment).toBe('mypurecloud.com');
+                // Localhost
+                query = '?gcHostOrigin=https://localhost:1337&gcTargetEnv=prod';
+                myClientApp = new ClientApp({
+                    gcHostOriginQueryParam: 'gcHostOrigin',
+                    gcTargetEnvQueryParam: 'gcTargetEnv'
+                });
+                expect(myClientApp.pcEnvironment).toBe('localhost');
```
justinaray, line 158: too bad we dont have access to the full `_pcEnv` prop to assert the value of pcAppOrigin.  But, i've never been much of a fan of exposing private members solely for the purpose of tests.

I dont see a great way of asserting it elsewhere either

src/ClientAppSpec.ts
```diff
@@ -139,6 +139,90 @@ export default describe('ClientApp', () => {
...
+                    });
+                }).toThrow();
+                // Localhost
+                query = '?gcHostOrigin=https://apps.mypurecloud.com&gcTargetEnv=invalid';
+                expect(() => {
+                    new ClientApp({
+                        gcHostOriginQueryParam: 'gcHostOrigin',
+                        gcTargetEnvQueryParam: 'gcTargetEnv'
+                    });
+                }).toThrow();
```
justinaray, line 177: can we add 1 more case here?  A valid origin and env but mis-matched?

Ex:
 query = '?gcHostOrigin=https://apps.mypurecloud.jp&gcTargetEnv=prod';

src/ClientAppSpec.ts
```diff
@@ -139,6 +139,90 @@ export default describe('ClientApp', () => {
...
+            });
+            it('should fail if the query params are badly configured or not set', () => {
+                const invalidQueryParamNames = [undefined, null, 3, [], {}, '', ' '];
+                invalidQueryParamNames.forEach((currInvalidParamName) => {
+                    expect(() => {
+                        new ClientApp({
+                            gcHostOriginQueryParam: 'gcHostOrigin',
+                            // @ts-expect-error
```
justinaray, line 212: nice usage of this opt-out here!

src/ClientAppSpec.ts
```diff
@@ -139,6 +139,90 @@ export default describe('ClientApp', () => {
...
+                    expect(() => {
+                        new ClientApp({
+                            gcTargetEnvQueryParam: 'gcTargetEnv',
+                            // @ts-expect-error
+                            gcHostOriginQueryParam: currInvalidParamName
+                        });
+                    }).toThrow();
+                });
+            });
+        });
```
justinaray, line 225: these tests are :fire:  nicely done

CONTRIBUTING.md
```diff
@@ -83,6 +83,17 @@ Specify the Genesys Cloud environment to target when host app is running on `loc
...
+``` json
+[
+    {
+        "env": "dev",
+        "publicDomainName": "some-domain-name.ext"
```
justinaray, line 92: super nit: `.tld` is more technically correct than `.ext`

src/utils/env.ts
```diff
@@ -26,6 +26,28 @@ const [defaultEnv] = environments.filter(env => env.region === DEFAULT_ENV_REGIO
 
 export const DEFAULT_PC_ENV = buildPcEnv(defaultEnv.publicDomainName);
 
+function findPcEnvironment(location: URL, targetEnv: string, parseEnvironment: EnvironmentParser): PcEnv|null {
```
justinaray, line 29: love the solution to pass in the parser!  :100:

src/utils/env.ts
```diff
@@ -26,6 +26,28 @@ const [defaultEnv] = environments.filter(env => env.region === DEFAULT_ENV_REGIO
...
+    const parsedEnv = parseEnvironment(location.origin, {});
+    if (parsedEnv && parsedEnv.name === targetEnv) {
+        return {
+            pcEnvTld: parsedEnv.publicDomainName,
+            pcAppOrigin: location.origin
+        };
+    } else {
+        for (const environment of __GC_DEV_EXTRA_ENVS__) {
+            const publicDomains = [environment.publicDomainName, ...(environment.publicDomainAliases || [])];
+            const matchingDomain = publicDomains.find((p) => location.hostname.endsWith(p));
```
justinaray, line 39: small nit here ... the env domains don't have a leading `.` so we have to add it here.

`location.hostname === p || location.hostname.endsWith(`.${p}`)`

otherwise, `apps.evilmypurecloud.com` would match `mypurecloud.com`

just lessons learned from a lot of unit testing parsing domains.  Granted, this is almost entirely for dev/testing, but, we might as well be safe.

src/utils/env.ts
```diff
@@ -65,3 +87,24 @@ export const lookupPcEnv = (pcEnvTld: string, lenient = false, envTlds: string[]
...
+        const hostLocation = new URL(url);
+        if (['localhost', '127.0.0.1'].includes(hostLocation.hostname)) {
+            return {
+                pcEnvTld: 'localhost',
+                pcAppOrigin: hostLocation.origin
+            };
+        } else {
+            return findPcEnvironment(hostLocation, targetEnv, parseEnvironment);
+        }
+    } catch {
```
justinaray, line 107: nice!  glad you wrapped this entire block because the `parse` method can throw if the url is invalid.

Which makes me realize we should add a test for that case so we're sure we return null (which eventually throws) if we're passed a malformed url for gcHostOrigin

src/utils/env.ts
```diff
@@ -65,3 +87,24 @@ export const lookupPcEnv = (pcEnvTld: string, lenient = false, envTlds: string[]
...
+
+/**
+ * Attempts to locate a PC environment corresponding to the provided url
+ * @param url A string representing the Genesys Cloud environment top-level domain to search for
+ * @returns A Genesys Cloud environment object if found; null otherwise.
+ */
+export const lookupGcEnv = (url: string, targetEnv: string, parseEnvironment: EnvironmentParser = parse): PcEnv|null => {
+    try {
+        const hostLocation = new URL(url);
+        if (['localhost', '127.0.0.1'].includes(hostLocation.hostname)) {
```
justinaray, line 99: hmm ... wonder if we should validate the targetEnv in this case.  Probably makes sense to given that apps will likely use that env to set the client api for public api.

Maybe we should just add a validate/assert targetEnv util method that checks if it's valid (against sd-web envs and our env-specified envs) before we even bother asserting the url?

justinaray:
@cadgerfeast  ... love the solution you came up with pass in the env parser so we can supplement with the env Envs.

tests look great and are passign.  And, I kicked the tires on the local server.

We're in the home stretch on this PR.  Only todos left are the small things I found on this last round of review:

* nit on custom env description text in CONTRIBUTING.md
* fix domain parsing with leading `.`
* validate targetEnv (in localhost case) 
* tests
	* add gcHostOrigin malformed url test
	* add localhost origin but invalid gcTargetEnv

src/ClientAppSpec.ts
justinaray: actually, you found a great solution by testing `lookupGcEnv` in the envSpec.  I rounded out a few tests so we could better assert the parsed values.  But, also think we should keep these here to test the constructor 👍

src/utils/env.ts
```diff
@@ -21,11 +21,38 @@ const PC_ENV_TLDS = environments
         return tlds;
     }, [] as string[])
     .concat(__PC_DEV_ENVS__);
+const GC_ENV_NAMES = new Set<string>([...environments, ...__GC_DEV_EXTRA_ENVS__].map((e) => e.name));
```
justinaray, line 24: @cadgerfeast ... tweaked this to use `name` instead of `env`

src/utils/env.ts
```diff
@@ -21,11 +21,38 @@ const PC_ENV_TLDS = environments
...
+    if (parsedEnv && parsedEnv.name === targetEnv) {
+        return {
+            pcEnvTld: parsedEnv.publicDomainName,
+            pcAppOrigin: location.origin
+        };
+    } else {
+        for (const environment of __GC_DEV_EXTRA_ENVS__) {
+            const publicDomains = [environment.publicDomainName, ...(environment.publicDomainAliases || [])];
+            const matchingDomain = publicDomains.find((p) => location.hostname === p || location.hostname.endsWith(`.${p}`));
+            if (environment.env === targetEnv && matchingDomain) {
```
justinaray, line 45: whoops, need `name` here as well

src/utils/env.ts
```diff
@@ -21,11 +21,38 @@ const PC_ENV_TLDS = environments
...
+    if (parsedEnv && parsedEnv.name === targetEnv) {
+        return {
+            pcEnvTld: parsedEnv.publicDomainName,
+            pcAppOrigin: location.origin
+        };
+    } else {
+        for (const environment of __GC_DEV_EXTRA_ENVS__) {
+            const publicDomains = [environment.publicDomainName, ...(environment.publicDomainAliases || [])];
+            const matchingDomain = publicDomains.find((p) => location.hostname === p || location.hostname.endsWith(`.${p}`));
+            if (matchingDomain && environment.name === targetEnv) {
```
justinaray, line 45: @cadgerfeast also switched this to `name`

justinaray (approved):
Great work @cadgerfeast ... let's :shipit:

PR #147, opened 2023-01-19
Support of gcHostOrigin-gcTargetEnv

cadgerfeast:
Adds support for `gcHostOrigin`/`gcTargetEnv` direct values in **ClientApp** constructor.

It also updates the remaining samples to use the new `gcHostOriginQueryParam`/`gcTargetEnvQueryParam` or `gcHostOrigin`/`gcTargetEnv` combination.


src/index.ts
```diff
@@ -151,30 +159,53 @@ class ClientApp {
...
-     *   pcEnvironmentQueryParam: 'pcEnvironment'
+     *   gcHostOriginQueryParam: 'gcHostOrigin',
+     *   gcTargetEnvQueryParam: 'gcTargetEnv'
      * });
      * ```
      *
      * @param cfg - Runtime config of the client
      */
     constructor(cfg: {
-        /** Name of a query param to auto parse into the pcEnvironment; Must be valid and a single param.  Best Practice. */
```
justinaray, line 161: nice catch!

src/index.ts
```diff
@@ -151,30 +159,53 @@ class ClientApp {
...
+         */
         gcTargetEnvQueryParam?: string;
+        /**
+         * The GC host origin (e.g. https://apps.mypurecloud.com, https://apps.mypurecloud.au).
+         * Must be a valid origin and used with gcTargetEnv.
+         */
+        gcHostOrigin?: string;
+        /**
+         * The GC env target (e.g. prod, prod-euw2).
+         * Must be a valid origin and used with gcHostOrigin.
```
justinaray, line 195: nit, `Must be a valid environment name`

examples/utils/oauth.js
```diff
@@ -1,5 +1,5 @@
 // Authenticate with Genesys Cloud
-function authenticate(client, pcEnvironment) {
+function authenticate(client, pcEnvironment, state) {
```
justinaray, line 2: Just need to assert all the calls are updated here.
Also, need to make sure we deploy these utils in all instances.  I do think it's the case, but, want to confirm.

examples/agentInteractionsDemo/app.js
```diff
@@ -152,19 +152,27 @@ new Vue({
...
-            });
-            Vue.prototype.$clientApp = clientApp;
+            if (queryParameters.gcHostOrigin && queryParameters.gcTargetEnv) {
+                clientApp = new window.purecloud.apps.ClientApp({
+                    gcHostOrigin: queryParameters.gcHostOrigin,
+                    gcTargetEnv: queryParameters.gcTargetEnv
+                });
+                pcEnvironment = clientApp.pcEnvironment;
+                Vue.prototype.$clientApp = clientApp;
+            } else if (pcEnvironment) {
```
justinaray, line 169: eventually, we can kill off support for pcEnvironment in the demo apps.  But, smart to leave it in for now.  Until we update all config using them.

examples/directoryNavigationDemo.html
```diff
@@ -272,32 +279,6 @@ <h2>Groups</h2>
                 }
             }
 
-            /**
-             * Determine the embedding Genesys Cloud environment seeded on the query string or
-             * being returned through the OAuth2 Implicit grant state hash param.
-             *
-             * @returns A string indicating the embedding PC env (e.g. mypurecloud.com, mypurecloud.jp); otherwise, null.
-             */
-            function getEmbeddingPCEnv() {
```
justinaray, line 281: thanks for cleaning up all these duplicated utils.  Years of tech debt in the making.

examples/externalContactsDemo/script.js
```diff
@@ -105,7 +109,7 @@ document.addEventListener('DOMContentLoaded', function () {
                         }
                     }
                 }).catch(err => {
-                    setErrorState(err);
+                    setErrorState(err.message);
```
justinaray, line 112: was this failing or just something you spotted?  Seems like the right change to make but tough to know in js and a catch block where err is `any`.

Dont mind the change ... just curious.

examples/utils/oauth.js
justinaray: lgtm!

examples/externalContactsDemo/script.js
cadgerfeast: When I tested the sample and got the permission error (missing app scope), the message shown in the progress bar was `[object Object]`, so I just made the change to show the real message sent by API.

cadgerfeast:
Hey @justinaray ,
I fixed the documentation issue and tested every sample on `2.6.0-alpha` branch with local and dev `web directory`.
Couldn't find any issue.

PR #148, opened 2023-01-25
Remove localhost target env check

cadgerfeast:
This env check is only being used when running `localhost` as it's being done be the `EnvironmentParser`.
Though when the `EnvironmentParser` is defined, the `isKnownEnvName` doesn't know about the list of environments.

### DIFF
--- a/src/utils/env.ts
+++ b/src/utils/env.ts
@@ -1,5 +1,5 @@
 /* eslint-disable no-console */
-import { EnvironmentFilters, getEnvironments } from 'genesys-cloud-service-discovery-web';
+import { getEnvironments } from 'genesys-cloud-service-discovery-web';
 
 export interface PcEnv {
     pcEnvTld: string;
@@ -12,8 +12,7 @@ const buildPcEnv = (tld: string): PcEnv => ({
 });
 
 const DEFAULT_ENV_REGION = 'us-east-1';
-const ENVIRONMENT_FILTERS: EnvironmentFilters = { env: ['prod', 'fedramp'], status: ['beta', 'stable'] };
-const environments = getEnvironments(ENVIRONMENT_FILTERS);
+const environments = getEnvironments({ env: ['prod', 'fedramp'], status: ['beta', 'stable'] });
 
 const PC_ENV_TLDS = environments
     .reduce((tlds, env) => {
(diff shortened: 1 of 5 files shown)


src/utils/env.ts
```diff
@@ -21,18 +22,25 @@ const PC_ENV_TLDS = environments
...
 }
 
-function findPcEnvironment(location: URL, targetEnv: string, parseEnvironment: EnvironmentParser): PcEnv|null {
-    const parsedEnv = parseEnvironment(location.origin, {});
+const matchesHostname = (hostname: string) => (domain: string) => {
+    return hostname === domain || hostname.endsWith(`.${domain}`);
+};
+
+function findPcEnvironment(location: URL, targetEnv: string, envs: Environment[]): PcEnv|null {
+    const parsedEnv = envs.find(({ publicDomainName, publicDomainAliases }) => {
```
justinaray, line 40: nit: since the logic is the same for both search blocks, should we setup the full env list first and then find?  Should be logically the same since we want to search for the default envs first.

```
[envs, ...__GC_DEV_EXTRA_ENVS__].find(...)
```

src/utils/env.ts
```diff
@@ -12,7 +12,8 @@ const buildPcEnv = (tld: string): PcEnv => ({
 });
 
 const DEFAULT_ENV_REGION = 'us-east-1';
-const environments = getEnvironments({ env: ['prod', 'fedramp'], status: ['beta', 'stable'] });
+const ENVIRONMENT_FILTERS: EnvironmentFilters = { env: ['prod', 'fedramp'], status: ['beta', 'stable'] };
```
justinaray, line 15: nit: i dont see any other usages of ENVIRONMENT_FILTERS or the EnvironmentFilters type.  If you moved the filters back inline into `getEnvironments`, then ts will infer the type and you wouldn't need to import it.

not hurting anything of course.

src/utils/env.ts
```diff
@@ -41,7 +49,7 @@ function findPcEnvironment(location: URL, targetEnv: string, parseEnvironment: E
     } else {
         for (const environment of __GC_DEV_EXTRA_ENVS__) {
             const publicDomains = [environment.publicDomainName, ...(environment.publicDomainAliases || [])];
```
justinaray, line 51: nit: missed this before but the `|| {}` bit should be unnecessary since the publicDomainAliases type is string[].

Will be moot if we combine these two blocks into one. 

Also is a nit and nbd ;)

justinaray (approved):
lgtm ... only nits

src/utils/env.ts
cadgerfeast: Yes Justin, I added this one in the case that the developer doesn't add `publicDomainAliases` to his `CLIENT_APP_SDK_GC_EXTRA_ENVS` variable. I'll remove this one and document the field.

PR #149, opened 2023-01-27
Add Environment type for TypeScript export

cadgerfeast:
The Environment type is defined globally in `globals.d.ts` for  `__GC_DEV_EXTRA_ENVS__`, but during build, the type is not added to the `dist/index.d.ts`, making it invalid.
The purpose of this fix is to add a valid export for npm package typings.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -9,7 +9,7 @@
  * @since 1.0.0
  */
 
-import { Environment } from 'genesys-cloud-service-discovery-web';
+import type { Environment } from 'genesys-cloud-service-discovery-web';
 import * as queryString from 'query-string';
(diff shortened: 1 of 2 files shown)


src/index.ts
```diff
@@ -9,6 +9,7 @@
  * @since 1.0.0
  */
 
+import type { Environment } from 'genesys-cloud-service-discovery-web';
```
justinaray, line 12: oh, was wondering how either of these files were resolving the type without the import since it's used in the file.  Now, I see that it's defined in globals.d.ts.

I was worried that VS Code's `organize imports` task would remove the import in favor of the global, but I just tried it and it looks like the import still sticks around.

justinaray (approved):

PR #150, opened 2023-01-30
Add gcEnvironment getter to client app

cadgerfeast:
The purpose of this change is to add the `gcEnvironment` getter to client apps.
This should be released with `2.6.3`. I also updated the sdk doc to mention this one, but samples have not been updated.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -268,6 +268,18 @@ class ClientApp {
      * @since 1.0.0
      */
     get pcEnvironment() {
+        return this.gcEnvironment;
+    }
+    /**
+     * Returns the gcEnvironment (e.g. mypurecloud.com, mypurecloud.jp) if known; null otherwise.
+     * This value will be available if a valid Genesys Cloud Environment is provided, inferred, or
+     * defaulted from the config passed to this instance.
+     *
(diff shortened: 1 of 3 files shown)


src/index.ts
```diff
@@ -268,6 +268,18 @@ class ClientApp {
...
+    /**
+     * Returns the gcEnvironment (e.g. mypurecloud.com, mypurecloud.jp) if known; null otherwise.
+     * This value will be available if a valid Genesys Cloud Environment is provided, inferred, or
+     * defaulted from the config passed to this instance.
+     *
+     * @returns the valid Genesys Cloud environment; null if unknown.
+     *
+     * @since 2.6.3
+     */
+    get gcEnvironment() {
```
justinaray, line 282: Once we can deprecate pcEnvironment, I think we'll also deprecate the string version of this in favor of targetEnvironment property of type `Environment` or `RuntimeEnvironment`.  The string verison is a bit ambiguous; but, for now, we can't reliably generate that type from the pcEnvironment input, so, we'll leave that as a breaking change for v3 of this lib or for appsv2

justinaray (approved):

PR #151, opened 2023-01-31
Update examples to latest sdk versions

cadgerfeast:
The idea is to bump sdk version for `purecloud-platform-client-v2` and `purecloud-client-app-sdk` to latest versions before publishing to github pages.


justinaray (approved):
Great work.  Bumped up all the examples with the right version/hash and updated the client sdk.  Thanks!

PR #153, opened 2023-07-19
Call stopped lifecycle method after toast popup

cadgerfeast:


justinaray (approved):
Tested out in v1 as well and all looks good.

PR #156, opened 2023-12-12
ATEAM-1157: Fix snyk vulnerability

carolinepar:
Bumping query string so that the decode-uri-component gets bumped and no longer causes a high snyk vulnerability. I needed to update some type checking to avoid some type errors as well

### DIFF
--- a/package.json
+++ b/package.json
@@ -61,7 +61,7 @@
   "dependencies": {
     "@babel/runtime": "^7.10.3",
     "genesys-cloud-service-discovery-web": "3.0.242",
-    "query-string": "^5.0.0"
+    "query-string": "7.1.3"
   },
   "devDependencies": {
     "@babel/core": "^7.10.2",
--- a/package.json
+++ b/package.json
@@ -81,7 +81,6 @@
     "@types/gh-pages": "^3.0.0",
     "@types/jasmine": "^3.5.10",
     "@types/lodash": "^4.14.159",
-    "@types/query-string": "^5.0.0",
     "@types/yargs": "^15.0.5",
     "@typescript-eslint/eslint-plugin": "^3.2.0",
     "@typescript-eslint/parser": "^3.2.0",
--- a/src/index.ts
+++ b/src/index.ts
@@ -202,7 +202,11 @@ class ClientApp {
             if ('gcHostOriginQueryParam' in cfg || 'gcTargetEnvQueryParam' in cfg) {
                 this.assertNonEmptyString(cfg.gcHostOriginQueryParam, 'host origin query param name');
                 this.assertNonEmptyString(cfg.gcTargetEnvQueryParam, 'target env query param name');
-                this._pcEnv = this.lookupGcEnv(parsedQueryString[cfg.gcHostOriginQueryParam], parsedQueryString[cfg.gcTargetEnvQueryParam]);
+                const parsedGcHostOrigin = parsedQueryString[cfg.gcHostOriginQueryParam];
+                const parsedGcTargetEnv = parsedQueryString[cfg.gcTargetEnvQueryParam];
+                this.assertNonEmptyString(parsedGcHostOrigin, 'host origin parsed query param');
+                this.assertNonEmptyString(parsedGcTargetEnv, 'target env parsed query param');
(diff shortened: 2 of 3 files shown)


bo-carey (approved):

src/index.ts
```diff
@@ -202,7 +202,11 @@ class ClientApp {
             if ('gcHostOriginQueryParam' in cfg || 'gcTargetEnvQueryParam' in cfg) {
                 this.assertNonEmptyString(cfg.gcHostOriginQueryParam, 'host origin query param name');
                 this.assertNonEmptyString(cfg.gcTargetEnvQueryParam, 'target env query param name');
-                this._pcEnv = this.lookupGcEnv(parsedQueryString[cfg.gcHostOriginQueryParam], parsedQueryString[cfg.gcTargetEnvQueryParam]);
+                const parsedGcHostOrigin = parsedQueryString[cfg.gcHostOriginQueryParam];
+                const parsedGcTargetEnv = parsedQueryString[cfg.gcTargetEnvQueryParam];
+                this.assertNonEmptyString(parsedGcHostOrigin, 'host origin parsed query param');
+                this.assertNonEmptyString(parsedGcTargetEnv, 'target env parsed query param');
```
justinaray, line 208: where you able to find any changelog for query-string?  I dont see anything in the repo.  Since this involves a major version bump, it has potential to break.

was this failing at runtime or did you identify it as a breaking change?  fortunately for us, this does look like our only usage in the lib.  So, if we confirm this is still working correctly, we should be fine.

carolinepar: I also was unable to find a changelog, but you are right that this is our only use for this package and I've tested it to make sure it still works. Although, one of the changes was them switching to typescript, which is how I caught this. Previously, parsedQueryString was type any, now its type is queryString.ParsedQuery<string>, so object keys are enforced as strings.

package.json
```diff
@@ -61,7 +61,7 @@
   "dependencies": {
     "@babel/runtime": "^7.10.3",
     "genesys-cloud-service-discovery-web": "3.0.242",
-    "query-string": "^5.0.0"
+    "query-string": "7.1.3"
```
justinaray, line 64: Can you use the cli to remove `@types/query-string` from package and lockfile?  Looks like query-string now publishes it's own types so we shouldn't need to add the external package.

https://www.npmjs.com/package/@types/query-string

src/index.ts
```diff
@@ -202,7 +202,11 @@ class ClientApp {
             if ('gcHostOriginQueryParam' in cfg || 'gcTargetEnvQueryParam' in cfg) {
                 this.assertNonEmptyString(cfg.gcHostOriginQueryParam, 'host origin query param name');
```
justinaray, line 203: @carolinepar ... these are now duplicated below on 207 and 208.  That's good as they'll typeguard the variables before being passed to the lookup function.  But, I think we can remove these now, right?

src/index.ts
justinaray, line 208: Trying to get the assert to fail but not having any luck with: https://localhost:8443/externalContactsDemo/index.html?gcHostOrigin=https://apps.inindca.com&gcTargetEnv=dev&gcTargetEnv=test

carolinepar: I'm able to get it to fail with this, since it's getting a null gcHostOrigin
https://localhost:8443/toast.html?pcEnvironment=inindca.com&pcLangTag=en-us&pcUsePopupAuth=true&gcHostOrigin&gcTargetEnv=dev&gcLangTag=en-us&gcUsePopupAuth=true

justinaray: @carolinepar ... I did confirm that the assertion is thrown and handled by the demo app.  I might have just missed the catch and visualzed error in the status bar.  Been awhile since I tested this demo.

image

I also tested with gcTargetEnv=prod,prod-usw2 and gcTargetEnv=prod&gcTargetEnv=prod-usw2 and this branch matches the existing behavior. So, we should be good with the few small clean-up fixes!

src/index.ts
justinaray: nvmd ... just realized this is asserting the param name in the config.  not the value.  nothing to see here :face_palm:

justinaray (approved):
looks good @carolinepar ... :shipit:

PR #157, opened 2024-04-12
ATEAM-1240: AlertingApi toast docs bugfix

carolinepar:
We received some feedback on the showToastPopup docs, and in looking into it I saw a bug as well, so this pr fixes both of these things! 
Current doc: https://developer.genesys.cloud/platform/integrations/client-apps/sdk/alertingapi#showtoastpopup

The customer said that it was unclear what units the toast timeout are in and that they weren't sure of the possible values for the type parameters. Because we do Typescript magic in the file, some of that documentation gets lost and doesn't make it to the markdown. To fix this I explicitly added it to the documentation above the function. I am happy to workshop the wording and formatting, if y'all have any ideas to make things more clear and concise let me know! 

The bug is in the parameters table in the `options` row. We set a default value in the parameters of the function, but we have to type cast it which messes up the formatting. The type case is necessary because we set the default toast type to `'info'` but the Type Parameters could be set to `'error' | 'warn'` for example, which would cause a type error. To fix this I moved the default out of the parameters and set it in the function instead. 

I struggled with getting the docs to build locally, but I used the Developer Center markdown editor to test the changes out and it looks like it's displaying correctly! To see the page for yourself, you can pull this branch, run `npm run docs`, copy the contents of `dist/docs/alertingapi.md` and paste it here: 
https://apicentral.inindca.com/services/developer-center/editor 

I've included screenshots as well. These screenshots don't include the middle code blocks because they have not been changed at all. 

toastDocs1
toastDocs2

Note: The old way of getting the docs to build locally is correct until the Developer Center gets migrated to Yeti CMS.

### DIFF
--- a/src/modules/alerting.ts
+++ b/src/modules/alerting.ts
@@ -70,7 +70,7 @@ const defaultToastOptions: ToastOptions<ValidMessageType, number> = {
     timeout: 7,
     showCloseButton: true,
     markdownMessage: true
-} as unknown as ToastOptions<ValidMessageType, number>;
+};
 
 /**
  * Handles aspects of alerting and attention of this app with Genesys Cloud
@@ -146,23 +146,23 @@ class AlertingApi extends BaseApi {
         message: string,
         options?: ToastOptions<MessageType, Timeout>
     ) {
-        const ToastOptions = options || defaultToastOptions as ToastOptions<MessageType, Timeout>;
+        const toastOptions = options || defaultToastOptions;
         const messageParams = {
             title,
             message,
             type: 'info'
         };
 
-        if (ToastOptions && typeof ToastOptions === 'object') {
-            if (ToastOptions.type && typeof ToastOptions.type === 'string') {
-                const requestedType = ToastOptions.type.trim().toLowerCase();
+        if (toastOptions && typeof toastOptions === 'object') {
+            if (toastOptions.type && typeof toastOptions.type === 'string') {
+                const requestedType = toastOptions.type.trim().toLowerCase();
 
                 if (isValidMessageType(requestedType)) {
                     messageParams.type = requestedType;
                 }
             }
 
-            const validOptions = pick(ToastOptions, VALID_SUPPLEMENTAL_OPTIONS);
+            const validOptions = pick(toastOptions, VALID_SUPPLEMENTAL_OPTIONS);
             Object.assign(messageParams, validOptions);
         }
 
--- a/src/modules/alerting.ts
+++ b/src/modules/alerting.ts
@@ -65,6 +65,13 @@ type ToastCloseButtonConfig<Timeout extends number> = Timeout extends 0
         showCloseButton?: boolean
     }
 
+const defaultToastOptions: ToastOptions<ValidMessageType, number> = {
+    type: 'info',
+    timeout: 7,
+    showCloseButton: true,
+    markdownMessage: true
+} as unknown as ToastOptions<ValidMessageType, number>;
+
 /**
  * Handles aspects of alerting and attention of this app with Genesys Cloud
  *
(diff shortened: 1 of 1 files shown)


src/modules/alerting.ts
```diff
@@ -65,6 +65,13 @@ type ToastCloseButtonConfig<Timeout extends number> = Timeout extends 0
         showCloseButton?: boolean
     }
 
+const defaultToastOptions: ToastOptions<ValidMessageType, number> = {
+    type: 'info',
+    timeout: 7,
+    showCloseButton: true,
+    markdownMessage: true
+} as unknown as ToastOptions<ValidMessageType, number>;
```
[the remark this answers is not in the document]
carolinepar: Ah you're so right the assertions are good to go! Thanks for catching that!

src/modules/alerting.ts
```diff
@@ -82,6 +89,17 @@ class AlertingApi extends BaseApi {
      * Error toasts (`type: 'error'`) require manual dismissal and must be explictly specified with `showCloseButton: true`.
      * TypeScript users will also specify `timeout: 0` while JavaScript users can specify 0 or omit the prop entirely.
      * The `timeout` prop will be ignored regardless.
+     * 
+     * **Toast Options:**
+     * 
+     * Name | Type | Default | Description |
```
justinaray, line 95: is the existence of a Default column enough to convey that these props are optional?

src/modules/alerting.ts
```diff
@@ -126,35 +144,31 @@ class AlertingApi extends BaseApi {
     >(
         title: string,
         message: string,
-        options: ToastOptions<MessageType, Timeout> = {
-            type: 'info',
-            timeout: 7,
-            showCloseButton: true,
-            markdownMessage: true
-        } as unknown as ToastOptions<MessageType, Timeout>
```
justinaray, line 134: I'm not sure I understand the bug here.  is it a runtime bug or a bug in generating the docs?

it's true that the type assertion to unknown first is risky but it's possibly acceptable here since the default is a constant and we're not taking any liberties with user provided values.

the default should only be applied if options is undefined, so, while

```ts
myClient.alerting.showToastPopup<'info', number>('foo', 'bar');
```

is technically incorrect, the function won't hold the type params so it's not as bad as a typed class which uses the types in other aspects of the class.

This does erorr correctly:
```ts
api.showToastPopup<'info', number>('foo', 'bar', {
    type: 'error',
    showCloseButton: true,
    timeout: 0
});

And the function types can be inferred from the final argument

```ts
api.showToastPopup('foo', 'bar', {
    type: 'error',
    showCloseButton: true,
    timeout: 0
});
```

I'm not opposed to this change at all, just not understanding the bug aspect of it.  I'm also not sure if applying the default inside the function is any more correct except that it does remove the type assertions on the default param.  But, it's very possible I'm missing a nuance.

src/modules/alerting.ts
justinaray: looking at the few optional params we have, it does seem like the default value is an indicator of optional:

image
image

So, probably ok to have the column and ensure the default is specified or we put `-` or `none` for required attrs.  Otherwise, g2g as is and dont think we need a ton more to flag optional

src/modules/alerting.ts
justinaray: ah, i see.  seems like we're not escaping the values going into the md table and that's what causes the issue.  specifically the `|` char in the union type is probably doing it.  I wonder if there is something we could change in the doc generation so we wouldn't have to change the code?

good to know the issue isn't at runtime or with the types

carolinepar: There is no `|` in the parameters, it's looking like it's a whitespace issue instead. When I get rid of the new character lines in the options parameter it does format the table correctly, although it does keep the type assertions which don't feel like they belong in the docs. 
image
I'm not too sure there's much we could do about the whitespace in terms of doc generation but I can take a look into it if we want to go that path. Thoughts?

justinaray (approved):

